PR #1, opened 2017-04-11
add missing commas

chen23:
The documentation example of using App Svcs iApp has invalid JSON.


ssurenr:
This issue was moved to ssurenr/k8s-bigip-ctlr#1

PR #2, opened 2017-06-13
centos7 & rhel7 image additions

tchughesiv:
Some additional mods are expected around the release/runtime make process... likely the build scripts though. The images should work as submitted.


sjberman:
@tchughesiv Have you successfully built the rhel7 based image on a RHEL host? I am attempting to build the image on a RHEL host with the proper subscription and repos enabled, but the build fails, telling me this: 
`Error getting repository data for rhel-7-server-rpms, repository not found
This system is not registered with an entitlement server. You can use subscription-manager to register.`

sjberman:
In doing some further investigation, this may be due to Docker Enterprise Edition not being able to access the host machine's repositories. I don't have this problem when running an older version of Docker.

sjberman:
@tchughesiv Just realized the conversation was locked! You couldn't reply even if you wanted to, sorry about that.

tchughesiv:
Saylor,
Make sure you're using Red Hat's docker package for your rhel7 image builds... e.g.:
https://github.com/RHsyseng/container-rhel-examples/tree/master/starter#building-an-image-on-rhel

PR #222, opened 2017-06-29
Fix docs config example

sjberman:
The JSON data was incorrectly formatted, causing the config example to be incorrect.

Removed the '-' that was not needed.


michaeldayreads:
@sjberman, @bmarshall13 assisted with #226 to address this. Thanks for hopping on it.

PR #223, opened 2017-06-29
Configure IPv6 Adddresses properly

sjberman:
Problem: The config driver only handled IPv4 Addresses, which ended in the form
':port', whereas IPv6 Addresses end in the form '.port'.

Solution: Correctly configure IPv4 vs IPv6 addresses.

### DIFF
--- a/python/bigipconfigdriver.py
+++ b/python/bigipconfigdriver.py
@@ -18,6 +18,7 @@ from __future__ import absolute_import
 import argparse
 import fcntl
 import hashlib
+import ipaddress
 import json
 import logging
 import os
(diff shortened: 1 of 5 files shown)


python/bigipconfigdriver.py
```diff
@@ -375,14 +376,22 @@ def create_ltm_config_kubernetes(bigip, config):
                 f5_service['virtual_address'] = \
                     svc['virtualAddress']['bindAddr']
 
+                addr = svc['virtualAddress']['bindAddr']
+                port = svc['virtualAddress']['port']
+                destination = None
+                if isinstance(ipaddress.ip_address(addr),
```
f5yacobucci: The controller's schema will ensure this is a valid address, correct? I think we can rely on that but otherwise should probably catch ipaddress.AddressValueError.

sjberman: Yes, the schema does validation.

sjberman:
In response to the unit test suggestion: Unit tests currently only test that bigipconfigdriver functions correctly, not whether or not CCCL will handle what we give it. The only functionality I needed to add to _f5.py was the ability to delete IPv6 addresses from the Virtual Address List when changed. The functionality for deleting IPv4 addresses is currently verified in a system test. If the upgrade to CCCL causes that test to fail, then we know that CCCL needs to be able to handle deleting IPv4 and IPv6 addresses when they change.

PR #228, opened 2017-07-03
Create pool even if pool members are gone

### LINKED ISSUE #227: Pools without pool members should still be configured
With the recent Ingress implementation, the way configs are stored resulted in the controller not configuring a pool if there were no members. This is incorrect. Pools should still exists even if there are no members.

sjberman:
Problem: With the recent Ingress additions, we stopped creating Pools if
the pool members were gone. This is incorrect behavior.

Solution: Still create pools even if members are gone, and handle an empty
pool member array in the python driver. Correctly print a log message when
there are no pool members configured.

Fixes #227

### DIFF
--- a/python/bigipconfigdriver.py
+++ b/python/bigipconfigdriver.py
@@ -402,8 +402,8 @@ def create_ltm_config_kubernetes(bigip, config):
 
     # FIXME(garyr): CCCL presently expects pools slightly differently than
     # we get from the controller, so convert to the expected format here.
-    found_svc = False
     for pool in f5_pools:
+        found_svc = False
         new_pool = {}
         members = {}
         pname = pool['name']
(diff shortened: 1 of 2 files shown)


python/bigipconfigdriver.py
```diff
@@ -424,13 +424,14 @@ def create_ltm_config_kubernetes(bigip, config):
...
+                found_svc = True
+                for member in pool['poolMemberAddrs']:
+                    members.update({member: {
+                        'state': 'user-up',
+                        'session': 'user-enabled'
+                    }})
+            new_pool['members'] = members
         configuration['pools'].append(new_pool)
         if not found_svc:
             log.warning(
```
f5yacobucci: I don't think this is an exceptional case any more and a warning is too sensitive as it indicates something went wrong. Let's move this to info.

PR #229, opened 2017-07-05
Fixed backoff timer exception handling.

ryan-talley:
Problem:
 Current exception handling will cause the backoff timer to not be reset
 correctly when broken configs are fixed and then broken again.

Solution:
 Increment the backoff timer when exceptions occur with configs so now
 it is incremented if an exception occurs or if regenerate_config_f5
 fails.

affects-branches: master


ryan-talley:
closing this PR since I have another PR that addresses this issue and another issue with the controller.

PR #230, opened 2017-07-05
Fix backoff timer

ryan-talley:
Two commits to fix two issues with the controller:

1) Fix exception handling in _do_reset to fix incorrect back off timer reseting.
2) Fix back off timer from an event to a Timer so it does not block _do_reset looping.

### DIFF
--- a/python/bigipconfigdriver.py
+++ b/python/bigipconfigdriver.py
@@ -451,7 +451,8 @@ class ConfigHandler():
         self._thread = threading.Thread(target=self._do_reset)
         self._pending_reset = False
         self._stop = False
-        self._backoff_timer = 1
+        self._backoff_time = 1
+        self._backoff_timer = None
         self._max_backoff_time = 128
 
         self._interval = None
@@ -476,6 +477,8 @@ class ConfigHandler():
         self._stop = True
         self._condition.notify()
         self._condition.release()
+        if self._backoff_timer is not None:
+            self.cleanup_backoff()
 
     def notify_reset(self):
         self._condition.acquire()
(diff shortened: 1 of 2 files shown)


python/bigipconfigdriver.py
```diff
@@ -554,17 +554,16 @@ def handle_backoff(self):
...
                 True):
             self._interval.stop()
-        self.retry_backoff(self.notify_reset)
+        self.retry_backoff()
 
-    def retry_backoff(self, func):
+    def retry_backoff(self):
         """Add a backoff timer to retry in case of failure."""
-        e = threading.Event()
+        timer = threading.Timer(self._backoff_time, self.notify_reset)
```
f5yacobucci: Unless this timer is cancelled it'll fire. So let's say we had an error, waited, config change fixes the error, timer fires - we'll notify a reset needlessly.

f5yacobucci: Since this slightly changes behavior I'd be interested in an additional test for that.

1) Start backoff with error (can start it with a high-ish wait)
2) Fix config and trigger a notify
3) Verify timer is cancelled (if timer.finished)

python/bigipconfigdriver.py
```diff
@@ -548,10 +544,18 @@ def _do_reset(self):
                               time.time() - start_time)
                 except Exception:
                     log.exception('Unexpected error')
+                    self.handle_backoff()
```
f5yacobucci: Can we add a test for this case?

f5yacobucci:
Existing tests are failing with this change. Some of the tests expect to test the Event state, they'll probably need to test the Timer state.

python/bigipconfigdriver.py
ryan-talley: Yes I will go ahead and add that test and fix the other test cases.

python/bigipconfigdriver.py
ryan-talley: This test has been added and it has been leveraged to cover the case that you brought up below.

PR #237, opened 2017-07-11
Run controller as non-root

sjberman:
Problem: We don't want container processes to be run as root.

Solution: Add a user to the runtime container and run processes as this user.

### DIFF
--- a/build-tools/Dockerfile.runtime
+++ b/build-tools/Dockerfile.runtime
@@ -3,7 +3,8 @@ FROM python:2.7-alpine
 ENV APPPATH /app
 
 RUN mkdir -p "$APPPATH/bin" \
- && chmod -R 755 "$APPPATH"
+ && chmod -R 755 "$APPPATH" \
+ && adduser -D ctlr
  
 WORKDIR $APPPATH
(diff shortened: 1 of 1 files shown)


build-tools/Dockerfile.runtime
```diff
@@ -3,7 +3,8 @@ FROM python:2.7-alpine
 ENV APPPATH /app
 
 RUN mkdir -p "$APPPATH/bin" \
- && chmod -R 755 "$APPPATH"
+ && chmod -R 755 "$APPPATH" \
+ && adduser -D -s /bin/bash user
```
f5gary: 2 issues here. 1 we should use a better name than 'user' - perhaps 'ctlr'. 2 I'm pretty sure that that we don't have bash installed and even if we did we don't need any bash-specific functionality so it's better to use /bin/sh as the shell. Do we even need to specify a shell for this to work?

sjberman: I'll check whether or not we need the shell.

sjberman:
Yes, this works with the systests.

PR #241, opened 2017-07-14
Update version and path info for v1.1.0

amudukutore:
Problem:
The version and path info need to be updated

Solution:
Changed the version info to point to 1.1.0 and product docs
path to 1.1


docs/RELEASE-NOTES.rst
```diff
@@ -1,8 +1,8 @@
 Release Notes for K8S BIG-IP Controller
 ============================================
 
-v1.1.0-dev
-----------
+v1.1.0
```
jputrino: This can be `|release|`

PR #249, opened 2017-07-18
1.1 stable

michaeldayreads:
CI cherry picks from master.


docs/RELEASE-NOTES.rst
```diff
@@ -1,33 +1,34 @@
...
-^^^^^^^^^^^^^^^^^^^
+```````````````````
 
-* Virtual Servers can now be configured on the BIG-IP from Kubernetes Ingress resources.
-* Multiple SSL Profiles can be configured for a Virtual Server.
-* Watches all Kubernetes namespaces by default, or can watch a list of namespaces, or namespaces with a customer specified label. This addresses a previous limitation in v1.0.0.
-* Watches for Kubernetes annotation if virtual address not specified, enabling custom IPAM integration.
-* Creates detached pools if virtual server bind addresses not specified.
+* Creation of BIG-IP Virtual Servers from Kubernetes Ingress resources.
+* Configure multiple SSL Profiles for a BIG-IP Virtual Servers.
```
amudukutore: Given that we say "for a", I would expect this to say "Virtual Server" and not "Virtual Servers"

michaeldayreads: @amudukutore - as this is a set of cherry picks, should we open an issue?

jputrino: @michaeldayreads - seconded. That is a separate issue.

amudukutore: Yes - separate issue that needs to be fixed in master and cherry-picked into 1.1-stable.

PR #267, opened 2017-07-24
Add OpenShift dependencies to add support for Route objects

f5gary:
Problem:
The controller project needs to include its openshift dependencies so
Route objects can be supported in OpenShift.

Solution:
Added the necessary dependencies to support OpenShift route objects.
Updated the .f5license file to contain the correct entries for
attributions generation.

affects-branches: master


f5gary:
Added another commit with the code that uses the openshift modules.

PR #270, opened 2017-07-26
Refactor syncVirtualServer into smaller functions

f5gary:
Problem:
The syncVirtualServer() function is too large.

Solution:
I refactored the ConfigMap specific code into syncConfigMap() and the
Ingress specific code into syncIngresses(). The overall functionality
has not changed.

affects-branches: master


sjberman:
Looks good to me. The build failed due to some network error it appears.

PR #276, opened 2017-08-02
Integrate Kubernetes BIG-IP Controller to use CCCL and CCCL Schema

edarzins:
Changes to the k8s-bigip-ctlr:
- Creation of a CCCL instance for each managed partition
- Modified create_ltm_config_kubernetes() to create CCCL-compliant schema
- Converted python unit tests

Note: This is an interim step. Currently, the creation of a BIG-IP configuration
occurs in two pieces: the golang frontend creates a config file from the
relevant configmaps and service descriptions, then the python backend performs
another conversion before passing the desired config to CCCL. The goal is to
create the CCCL-compliant BIG-IP config in one step within the golang frontend
of k8s-bigip-ctlr. Most importantly, this commit removes the dependency upon the
legacy f5-cccl/f5_cccl/_f5.py implemention.

### DIFF
--- a/python/bigipconfigdriver.py
+++ b/python/bigipconfigdriver.py
@@ -33,10 +33,9 @@ import urllib
 import pyinotify
 
 from urlparse import urlparse
-from f5_cccl._f5 import CloudBigIP, get_protocol, has_partition, log_sequence
-from f5_cccl.common import extract_partition_and_name, ipv4_to_mac,\
-    list_diff_exclusive, IPV4FormatError, PartitionNameError
 from f5.bigip import ManagementRoot
+from f5_cccl.api import F5CloudServiceManager
+from f5_cccl.exceptions import F5CcclError
 
 log = logging.getLogger(__name__)
 console = logging.StreamHandler()
@@ -44,6 +43,24 @@ console.setFormatter(
     logging.Formatter("[%(asctime)s %(name)s %(levelname)s] %(message)s"))
 root_logger = logging.getLogger()
 root_logger.addHandler(console)
+
+SCHEMA_PATH = "./src/f5-cccl/f5_cccl/schemas/cccl-api-schema.yml"
+
+
+class PartitionNameError(Exception):
+    """Exception type for F5 resource name."""
+
+    def __init__(self, msg):
+        """Create partition name exception object."""
+        Exception.__init__(self, msg)
+
+
+class IPV4FormatError(Exception):
+    """Exception type for improperly formatted IPv4 address."""
+
+    def __init__(self, msg):
+        """Create ipv4 format exception object."""
+        Exception.__init__(self, msg)
 
 
 class ResponseStatusFilter(logging.Filter):
@@ -65,12 +82,72 @@ root_logger.addFilter(ResponseStatusFilter())
 root_logger.addFilter(CertFilter())
 root_logger.addFilter(KeyFilter())
 
+
+def list_diff_exclusive(list1, list2):
+    """Return items found only in list1 or list2."""
+    return list(set(list1) ^ set(list2))
+
+
+def ipv4_to_mac(ip_str):
+    """Convert an IPV4 string to a fake MAC address."""
+    ip = ip_str.split('.')
+    if len(ip) != 4:
+        raise IPV4FormatError('Bad IPv4 address format specified for '
+                              'FDB record: {}'.format(ip_str))
+
+    return "0a:0a:%02x:%02x:%02x:%02x" % (
+        int(ip[0]), int(ip[1]), int(ip[2]), int(ip[3]))
+
+
+def extract_partition_and_name(f5_partition_name):
+    """Separate partition and name components for a Big-IP resource."""
+    parts = f5_partition_name.split('/')
+    count = len(parts)
+    if f5_partition_name[0] == '/' and count == 3:
+        # leading slash
+        partition = parts[1]
+        name = parts[2]
+    elif f5_partition_name[0] != '/' and count == 2:
+        # leading slash missing
+        partition = parts[0]
+        name = parts[1]
+    else:
+        raise PartitionNameError('Bad F5 resource name encountered: '
+                                 '{}'.format(f5_partition_name))
+    return partition, name
+
+
+def log_sequence(prefix, sequence_to_log):
+    """Helper function to log a sequence.
+
+    Dump a sequence to the logger, skip if it is empty
+
+    Args:
+        prefix: The prefix string to describe what's being logged
+        sequence_to_log: The sequence being logged
+    """
+    if sequence_to_log:
+        log.debug(prefix + ': %s', (', '.join(sequence_to_log)))
+
+
+def get_protocol(protocol):
+    """Return the protocol (tcp or udp)."""
+    if str(protocol).lower() == 'tcp':
+        return 'tcp'
+    if str(protocol).lower() == 'http':
+        return 'tcp'
+    if str(protocol).lower() == 'udp':
+        return 'udp'
+    else:
+        return None
+
+
 DEFAULT_LOG_LEVEL = logging.INFO
 DEFAULT_VERIFY_INTERVAL = 30.0
 
 
-class K8sCloudBigIP(CloudBigIP):
-    """K8sCloudBigIP class.
+class K8sCloudServiceManager():
+    """K8sCloudServiceManager class.
 
     Generates a configuration for a BigIP based upon the apps/tasks managed
     by services/pods/nodes in Kubernetes.
@@ -80,26 +157,32 @@ class K8sCloudBigIP(CloudBigIP):
       BigIP partition
     - For each backend (task, node, or pod), it creates a pool member and adds
       the member to the pool
-    - If the app has a Marathon Health Monitor configured, create a
-      corresponding health monitor for the BigIP pool member
     - Token-based authentication is used by specifying a token named 'tmos'.
       This will allow non-admin users to use the API (BIG-IP must configure
       the accounts with proper permissions, for either local or remote auth).
 
     Args:
-        hostname: IP address of BIG-IP
-        username: BIG-IP username
-        password: BIG-IP password
-        partitions: List of BIG-IP partitions to manage
+        bigip: ManagementRoot object
+        partition: BIG-IP partition to manage
+        schema_path: Path to the CCCL schema
     """
 
-    def __init__(self, hostname, port, username, password, partitions,
-                 manage_types):
-        """Initialize the K8sCloudBigIP object."""
-        super(K8sCloudBigIP, self).__init__(hostname, port, username,
-                                            password, partitions,
-                                            token="tmos",
-                                            manage_types=manage_types)
+    def __init__(self, bigip, partition, schema_path):
+        """Initialize the K8sCloudServiceManager object."""
+        self._mgmt_root = bigip
+        self._cccl = F5CloudServiceManager(
+            bigip,
+            partition,
+            prefix="",
+            schema_path=schema_path)
+
+    def mgmt_root(self):
+        """ Return the BIG-IP ManagementRoot object"""
+        return self._mgmt_root
+
+    def get_partition(self):
+        """ Return the managed partition."""
+        return self._cccl.get_partition()
 
     def _apply_config(self, config):
         """Apply the configuration to the BIG-IP.
@@ -107,10 +190,13 @@ class K8sCloudBigIP(CloudBigIP):
         Args:
             config: BIG-IP config dict
         """
+        incomplete = 0
         if 'ltm' in config:
-            CloudBigIP._apply_config(self, config['ltm'])
+            incomplete = self._cccl.apply_config(config['ltm'])
         if 'network' in config:
-            self._apply_network_config(config['network'])
+            incomplete += self._apply_network_config(config['network'])
+
+        return incomplete
 
     def _apply_network_config(self, config):
         """Apply the network configuration to the BIG-IP.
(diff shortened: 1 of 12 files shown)


python/bigipconfigdriver.py
```diff
@@ -80,38 +178,38 @@ class K8sCloudBigIP(CloudBigIP):
...
     """
 
-    def __init__(self, hostname, port, username, password, partitions,
-                 manage_types):
-        """Initialize the K8sCloudBigIP object."""
-        super(K8sCloudBigIP, self).__init__(hostname, port, username,
-                                            password, partitions,
-                                            token="tmos",
-                                            manage_types=manage_types)
+    def __init__(self, bigip, partition, schema_path, manage_types):
```
f5gary: manage_types doesn't seem to be passed down to cccl. We need this to enable policies, and in the near future iRules.

python/bigipconfigdriver.py
```diff
@@ -401,31 +621,31 @@ def create_ltm_config_kubernetes(bigip, config):
...
         new_pool = {}
-        members = {}
+        members = []
         pname = pool['name']
         new_pool['name'] = pname
-        monitors = None
         if 'monitor' in pool and pool['monitor']:
-            monitors = ' and '.join(pool['monitor'])
-        new_pool['monitor'] = monitors
+            log.debug("HERE")
```
f5gary: Leftover debugging

f5gary:
There is a lot to see here, please schedule an in-person review.

python/bigipconfigdriver.py
```diff
@@ -609,51 +845,60 @@ def _do_reset(self):
...
-                    if self._bigip.regenerate_config_f5(cfg):
+                    incomplete = 0
+                    for cccl in self._cccls:
+                        cfg = create_config_kubernetes(cccl, config)
+                        try:
+                            incomplete += cccl._apply_config(cfg)
+
+                            # Manually delete custom profiles (if needed)
+                            if customProfiles:
+                                _delete_client_ssl_profiles(
```
andrewjjenkins: All of our special custom configurators (like _delete_client_ssl_profiles, _create_client_ssl_profiles and so on) should have the same error behavior as cccl._apply_config(cfg) .  I think this looks like:

- For any serious errors (failure to connect with CCCL, broken protocol, violated schema), throw a CCCL exception
- For any problem actually implementing a config on BIG-IP, return an incomplete counter.

Then our try/catch/Incomplete/backoff logic is the same for everything

python/bigipconfigdriver.py
```diff
@@ -609,51 +845,60 @@ def _do_reset(self):
...
+                                profile)
                             customProfiles = True
 
-                    cfg = create_config_kubernetes(self._bigip, config)
-                    if self._bigip.regenerate_config_f5(cfg):
+                    incomplete = 0
+                    for cccl in self._cccls:
+                        cfg = create_config_kubernetes(cccl, config)
+                        try:
+                            incomplete += cccl._apply_config(cfg)
```
andrewjjenkins: This is confusing because it is calling the _apply_config() from earlier in this file (because cccl in this case isn't a real cccl, it's our derived extension of cccl).

Erik, I think you said you'd look at changing this to has-a-cccl instead of is-a-cccl

python/bigipconfigdriver.py
```diff
@@ -609,51 +845,60 @@ def _do_reset(self):
...
+
+                            # Manually delete custom profiles (if needed)
+                            if customProfiles:
+                                _delete_client_ssl_profiles(
+                                    cccl, config['resources'])
+
+                        except F5CcclError as e:
+                            log.error("CCCL Error: %s", e.msg)
```
andrewjjenkins: This won't trigger backoff.  We have three error options in this program:

1. Error, trigger backoff and retry.
2. Error, log and do nothing (until next update).
3. Error, crash.

This is "class 2".  I think we should consider eliminating "class 2" and treat errors as "class 1" or "class 3".  What errors get us into this case and why should we not backoff or crash?

python/bigipconfigdriver.py
```diff
@@ -609,51 +845,60 @@ def _do_reset(self):
...
+                        test_data['Total_Backends'] = backend_count
+                        test_data['Time'] = time.time()
+                        json_data = json.dumps(test_data)
+                        log.info('SCALE_PERF: Test data: %s',
+                                 json_data)
 
                     log.debug('updating tasks finished, took %s seconds',
                               time.time() - start_time)
+
                 except Exception:
```
andrewjjenkins: This is scary.  I think you shouldn't catch all exceptions, you should catch expected runtime exceptions (and handle them).

python/bigipconfigdriver.py
```diff
@@ -960,23 +1204,36 @@ def main():
...
+        bigip = ManagementRoot(
+            host,
+            config['bigip']['username'],
+            config['bigip']['password'],
+            port=port,
+            token="tmos")
+
+        # Management for the BIG-IP partitions
+        cccls = []
+        for partition in config['bigip']['partitions']:
```
andrewjjenkins: This has a risk that's only a problem if someone breaks a contract, but here it is:

We only create cccls for partitions declared the first time we read the config file.  The contract between k8s-bigip-ctlr (go code) and this file is that you can't change that list over time - you have to restart.  However, IMO it's worth verifying that contract is being obeyed since these are two programs in two different languages (there's a fair amount of distance for something to get dropped and this contract to be violated).

We talked about how this risk already exists, and so we decided to fix this later (once we've hoisted a lot of this into the go code).  That's fine with me.

python/bigipconfigdriver.py
edarzins: Removed.

python/bigipconfigdriver.py
```diff
@@ -45,6 +44,24 @@
 root_logger = logging.getLogger()
 root_logger.addHandler(console)
 
+SCHEMA_PATH = "./src/f5-cccl/f5_cccl/schemas/cccl-api-schema.yml"
```
edarzins: Installed via 'pip install -e'

python/bigipconfigdriver.py
```diff
@@ -80,38 +178,38 @@ class K8sCloudBigIP(CloudBigIP):
...
 
         Args:
             config: BIG-IP config dict
         """
+        incomplete = 0
         if 'ltm' in config:
-            CloudBigIP._apply_config(self, config['ltm'])
+            incomplete = self.apply_config(config['ltm'])
         if 'network' in config:
             self._apply_network_config(config['network'])
```
edarzins: Agree

python/bigipconfigdriver.py
```diff
@@ -155,7 +253,8 @@ def get_vxlan_tunnel(self, vxlan_name):
             vxlan_name: Name of the vxlan tunnel
         """
         partition, name = extract_partition_and_name(vxlan_name)
-        vxlan_tunnel = self.net.fdb.tunnels.tunnel.load(
+        mgmt = self._bigip_proxy.mgmt_root()
```
edarzins: Agree

python/bigipconfigdriver.py
```diff
@@ -267,7 +366,7 @@ def __init__(self, msg):
         Exception.__init__(self, msg)
 
 
-def create_config_kubernetes(bigip, config):
+def create_config_kubernetes(cccl, config):
```
edarzins: Agree

python/bigipconfigdriver.py
```diff
@@ -306,34 +405,128 @@ def append_ssl_profile(profiles, profName):
                          'name': profile[1]})
 
 
-def create_ltm_config_kubernetes(bigip, config):
+def iapp_build_definition(config, members):
```
edarzins: I will create an issue in CCCL to address this.

python/bigipconfigdriver.py
```diff
@@ -435,35 +655,51 @@ def create_ltm_config_kubernetes(bigip, config):
...
             log.info(
                 'Pool "{}" has service "{}", which is empty - '
                 'configuring 0 pool members.'.format(
                     pname, pool['serviceName']))
 
+    # Append the protocol to the monitor names to keep them unique
+    for monitor in configuration.get('monitors', []):
+        old_name = monitor['name']
+        old_path = '/' + cccl.get_partition() + '/' + old_name
+        suffix = '_0_' + monitor['type']
```
edarzins: Agree

python/bigipconfigdriver.py
```diff
@@ -609,51 +845,60 @@ def _do_reset(self):
                     # Manually create custom profiles;CCCL doesn't yet do this
                     if 'customProfiles' in config['resources']:
                         for profile in config['resources']['customProfiles']:
-                            _create_client_ssl_profile(self._bigip, profile)
+                            _create_client_ssl_profile(
```
edarzins: Agree

python/bigipconfigdriver.py
edarzins: Agree

python/bigipconfigdriver.py
edarzins: Agree

python/bigipconfigdriver.py
edarzins: Agree

python/bigipconfigdriver.py
edarzins: Agree

python/bigipconfigdriver.py
edarzins: Agree

python/bigipconfigdriver.py
f5gary: I have looked into this in more detail, and while CCCL still has some of the manage_types code in it still, it has no effect. Therefore in this file manage_types should be removed from this code altogether (and eventually from CCCL).

edarzins: OK

python/bigipconfigdriver.py
edarzins: https://github.com/f5devcentral/f5-cccl/issues/139

python/bigipconfigdriver.py
```diff
@@ -596,67 +846,83 @@ def _do_reset(self):
...
-                        if (self._interval and self._interval.is_running() is
-                                False):
-                            self._interval.start()
-                        self._backoff_time = 1
-                        if self._backoff_timer is not None:
-                            self.cleanup_backoff()
+                start_time = time.time()
+
+                config = _parse_config(self._config_file)
+                if 'resources' not in config:
```
andrewjjenkins: If you have no resources, but do have an Openshift SDN config (config['openshift-sdn']), I think you do want to configure.

As well, if you have no resources, but are supposed to be in charge of 2 partitions, I think you do want to "configure" them (to remove everything from inside of them).

If that's right, I don't think you want this "continue"

python/bigipconfigdriver.py
```diff
@@ -596,67 +846,83 @@ def _do_reset(self):
...
+                    continue
+                verify_interval, _ = _handle_global_config(config)
+                _handle_openshift_sdn_config(config)
+                self.set_interval_timer(verify_interval)
+
+                incomplete = 0
+
+                for mgr in self._managers:
+                    cfg = create_config_kubernetes(mgr.get_partition(),
```
andrewjjenkins: Thanks!  It's a lot easier to understand the control flow (and that create_config_kubernetes isn't actually touching the BIG-IP) when I see the minimal arguments you added here.

I know it wasn't a change to the program behavior at all but it is a big help in readability to someone like me who understands what the program is supposed to do but doesn't spend very much time in the details.

python/bigipconfigdriver.py
```diff
@@ -596,67 +846,83 @@ def _do_reset(self):
...
+                                                   config)
+
+                    try:
+                        incomplete += mgr._apply_config(cfg)
+
+                        # Manually create custom profiles;
```
andrewjjenkins: This will now happen after mgr._apply_config(cfg), but I think it happened before mgr._apply_config(cfg) before your PR.

Does this matter?  I think it might since some of the virtual servers that mgr._apply_config() is trying to configure refer to custom profiles that you might not have created yet.  If you're not sure, suggest you do it like this:

1. Manually create custom profiles
2. mgr._apply_config(cfg)
3. Manually delete custom profiles (if needed)

python/bigipconfigdriver.py
```diff
@@ -507,32 +739,50 @@ def _create_client_ssl_profile(bigip, profile):
...
+                try:
                     prof.delete()
+                except Exception as err:
+                    log.error("Error deleting SSL profile: %s" % err.message)
+                    incomplete += 1
+
+    return incomplete
 
 
 class ConfigHandler():
```
andrewjjenkins: Cool, I think the has-a approach worked out well (at least it's more clear to me)

edarzins: Agree

python/bigipconfigdriver.py
edarzins: Agree

python/bigipconfigdriver.py
edarzins: Yes, that became obvious as I refactored it.

python/bigipconfigdriver.py
f5gary: No, please leave the continue in. This is the case where the controller has just (re)started and hasn't received all of the objects from the apiserver yet. Without the continue, you will wipe out the BIG-IP configuration before we know what the configuration is.

edarzins: Good catch! This was existing code and I agree that it's a bug!

f5gary: It's not a bug, it's a fix for a bug where we prematurely wiped the BIG-IP before we knew what the configuration should be.

edarzins: Upon further review...no resources could mean that the ctlr is not ready yet. Keep it as-is with a comment.

PR #277, opened 2017-08-03
Route support for OpenShift

sjberman:
Add Route object support for OpenShift. This allows users to create
Route objects that are translated into BIG-IP resources and policies
for routing traffic.

Add unit test utilities and tests for managing Route objects.
Also moved an Ingress unit test to its proper location in resourceConfig_test,
which is where a similar unit test for Routes lives.

### DIFF
--- a/cmd/k8s-bigip-ctlr/main.go
+++ b/cmd/k8s-bigip-ctlr/main.go
@@ -65,6 +65,7 @@ var (
 	bigIPFlags        *pflag.FlagSet
 	kubeFlags         *pflag.FlagSet
 	openshiftSDNFlags *pflag.FlagSet
+	osRouteFlags      *pflag.FlagSet
 
 	pythonBaseDir    *string
 	logLevel         *string
@@ -87,6 +88,10 @@ var (
 	openshiftSDNMode string
 	openshiftSDNName *string
 
+	routeVserverAddr       *string
+	routeDefaultServerCert *string
+	routeLabel             *string
+
 	// package variables
 	isNodePort         bool
 	watchAllNamespaces bool
@@ -98,6 +103,7 @@ func _init() {
 	bigIPFlags = pflag.NewFlagSet("BigIP", pflag.ContinueOnError)
 	kubeFlags = pflag.NewFlagSet("Kubernetes", pflag.ContinueOnError)
 	openshiftSDNFlags = pflag.NewFlagSet("Openshift SDN", pflag.ContinueOnError)
+	osRouteFlags = pflag.NewFlagSet("OpenShift Routes", pflag.ContinueOnError)
 
 	// Global flags
 	pythonBaseDir = globalFlags.String("python-basedir", "/app/python",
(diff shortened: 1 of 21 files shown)


cmd/k8s-bigip-ctlr/main.go
```diff
@@ -419,7 +446,8 @@ func main() {
 		log.Fatalf("error connecting to the client: %v", err)
 	}
 	if *manageRoutes {
-		appMgrParms.RouteClientV1, err = routeclient.New(config)
+		rclient, err := routeclient.New(config)
```
f5gary: Now that we know what the actual command line options for routes are, we should remove manageRoutes and use the routeVserverAddr as the flag to turn on route support. Also, all of the new options should be 'hidden' until we're done with the feature.

sjberman: Ok, I can hide the options. As of right now, not including the routeVserverAddr will enable pool only mode. Do we want to keep that functionality for Routes?

pkg/appmanager/appManager.go
```diff
@@ -90,20 +89,30 @@ type Manager struct {
...
+	// Route configurations
+	routeConfig RouteConfig
 }
 
 // Struct to allow NewManager to receive all or only specific parameters.
 type Params struct {
 	KubeClient      kubernetes.Interface
 	restClient      rest.Interface // package local for unit testing only
-	RouteClientV1   *routeclient.Client
+	RouteClientV1   rest.Interface
```
f5gary: This may cause difficulty down the road when we need to send events on a Route object. As I recall, the routeclient.Client was an object whose only member was the rest.Interface, but it also satisfied the interface for sending updates back to k8s.

sjberman: Right now we use the KubeClient to send updates back to k8s for Ingress/ConfigMaps. Using this rest.Interface allows the unit tests to be able to mock it properly. Maybe there is a way to also pull in routeclient.Client as a separate var for updating Routes.

pkg/appmanager/appManager.go
```diff
@@ -1225,23 +1323,29 @@ func (appMgr *Manager) handleConfigForType(
...
 		}
-		rsMap[pool.ServicePort] = cfgList
 	}
 
 	if _, ok := svcPortMap[pool.ServicePort]; !ok {
 		log.Debugf("Process Service delete - name: %v namespace: %v",
 			pool.ServiceName, svcKey.Namespace)
-		log.Infof("Ports for service '%v' have not been found.",
-			pool.ServiceName)
+		log.Infof("Port '%v' for service '%v' have not been found.",
```
f5gary: Grammar nit: "Port '%v' for service '%v' was not found."

pkg/appmanager/appManager.go
```diff
@@ -944,14 +960,95 @@ func (appMgr *Manager) syncIngresses(
...
+			if cfg.Virtual.Partition == rsCfg.Virtual.Partition &&
+				!reflect.DeepEqual(cfg, rsCfg) {
+				cfg = &rsCfg
+				appMgr.resources.Assign(keys[i], cfg.Virtual.VirtualServerName, cfg)
+			}
+		}
+		appMgr.resources.Unlock()
+
+		// TLS Cert/Key
+		if nil != route.Spec.TLS && len(route.Spec.TLS.Termination) > 0 {
```
f5gary: Should probably check for Termination == "edge" here rather than the length chck, as the logic will be different for the other values ("passthrough" and "reencrypt").

sjberman: Something like this in the F5Router? https://github.com/openshift/origin/blob/master/pkg/router/f5/plugin.go#L413

pkg/appmanager/appManager.go
f5gary: That's likely how we will do it.

pkg/appmanager/appManager.go
f5gary: Yes

cmd/k8s-bigip-ctlr/main.go
f5gary: Perhaps. I guess leave manageRoutes as-is for now until we work through all the details, etc.

f5gary:
This looks good to me besides the comments I made. I see the CI build is failing, which I assume will be fixed once you can finalize the changes in bigipconfigdriver.py

PR #283, opened 2017-08-17
RHEL base image for OpenShift

sjberman:
Problem: Need to use a RHEL7 base image for the OpenShift controller. The Kubernetes
version will continue to use Alpine as a base.

Solution: Add Dockerfiles for RHEL7 and adjust the build steps to support both Alpine
and RHEL7 images.


build-tools/Dockerfile.rhel7.runtime
```diff
@@ -0,0 +1,51 @@
+FROM registry.access.redhat.com/rhel-atomic
+
+LABEL name="f5networks/k8s-bigip-ctlr" \
+      vendor="F5 Networks" \
+      version="1.2" \
+      release="1" \
```
bmarshall13: The version should be passed in with the docker build `--label` option, so that the version can be recorded in one place.

build-tools/_build-lib.sh
```diff
@@ -66,7 +66,7 @@ go_install () {
   (
     export GOBIN="$BUILDDIR/bin"
     echodo cd "$BUILDDIR"
-    echodo go install $BUILD_VARIANT_FLAGS -v "$pkg"
+    echodo go install $BUILD_VARIANT_FLAGS -v -ldflags="-s -w" "$pkg"
```
bmarshall13: I don't think we want to strip symbols. At least not as part of this PR.

build-tools/entrypoint.builder.sh
```diff
@@ -7,7 +7,15 @@
 USER_ID=${LOCAL_USER_ID:-9001}
 
 echo "Starting with UID : $USER_ID"
-adduser -D -s /bin/bash -u $USER_ID user
 export HOME=/home/user
 
-exec /sbin/su-exec user "$@"
+if [ -v GOSU_VERSION ]; then
```
bmarshall13: `if [ -x /sbin/su-exec ]` would make the check more readable. Otherwise you need to look at the Dockerfile to see what this variable means

Makefile
```diff
@@ -5,7 +5,14 @@ GOOS     = $(shell go env GOOS)
...
 
-GO_BUILD_FLAGS=-v
+GO_BUILD_FLAGS=-v -ldflags="-s -w"
+
+# Allow users to pass in OS build options
+ifeq ($(TARGET),rhel7)
+	OS := rhel7
+else
+	OS := alpine
+endif
```
bmarshall13: This variable name should be more distinctive, and the same throughout. Maybe 'BASE_OS'.

Instead of an if here, just provide a default: `BASE_OS ?= alpine`  Then `make BASE_OS=rhel7 ...` works as expected.

Jenkinsfile
```diff
@@ -0,0 +1,37 @@
...
+  agent {
+    label "rhel7"
+  }
+  stages {
+    stage("build") {
+      steps {
+        sh '''#!/usr/bin/env bash
+          DOCKER_NAMESPACE="docker-registry.pdbld.f5net.com/velcro"
+          BASE_PUSH_TARGET="$DOCKER_NAMESPACE/k8s-bigip-ctlr"
+          export IMG_TAG="${BASE_PUSH_TARGET}:${GIT_COMMIT}-rhel7"
```
bmarshall13: Use $BASE_OS instead of hard coding rhel7. (And other instances in this file).

build-tools/Dockerfile.rhel7.runtime
```diff
@@ -0,0 +1,49 @@
+FROM registry.access.redhat.com/rhel-atomic
+
+LABEL name="f5networks/k8s-bigip-ctlr" \
+      vendor="F5 Networks" \
+### Required labels above - recommended below
```
bmarshall13: Maybe remove this comment, or indicate that the version should be passed in as an arg to docker build.

PR #284, opened 2017-08-18
Add support for SSL passthrough termination

f5gary:
Problem:
The controller needs to support SSL passthrough termination to achieve
paritiy with the OpenShift F5 Router.

Solution:
SSL passthrough is achieved by using an iRule to look at the SNI
information to get the target server name. The iRule uses internal
data groups that contain information that allows the iRule to map the
server name to the correct pool and then pass the connection through
unencrypted to that pool. The iRule itself was copied from the
OpenShift F5 Router verbatim. To support all of this new functionality,
the controller was modified as follows:
1. Added support for the 'iRules' section in the output configuration.
2. Updated the Virtual struct to include an array of fully-qualified
   iRule names in the form of '/Partition/iRuleName'.
3. Added support for the 'internalDataGroups' section in the output
   configuration.
4. Create an up to date list for the mapping of server name to pool
   every time a route may have changed, which may trigger a new config
   to be written.
5. Added the appropriate handling of 'termination-type' ==
   'passthrough', which exposed some cases where type 'edge' was being
   incorrectly assumed.
6. Updated the bigipconfigdriver to pass the new iRule and internal
data group information through to CCCL.

Fixes VEL-1012

affects-branches: master

### DIFF
--- a/pkg/appmanager/routing.go
+++ b/pkg/appmanager/routing.go
@@ -316,6 +316,8 @@ when CLIENT_DATA {
 	return iRuleCode
 }
 
+// Update a specific datagroup for passthrough routes, indicating if
+// something had changed.
 func (appMgr *Manager) updatePassthroughRouteDataGroups(
 	partition string,
 	poolName string,
@@ -343,7 +345,8 @@ func (appMgr *Manager) updatePassthroughRouteDataGroups(
 	return changed, nil
 }
 
-func updateDataGroupForRoute(
+// Update a data group map based on a passthrough route object.
+func updateDataGroupForPassthroughRoute(
 	route *routeapi.Route,
 	partition string,
 	dgMap InternalDataGroupMap,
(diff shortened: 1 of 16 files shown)


pkg/appmanager/routing.go
```diff
@@ -201,3 +211,193 @@ func newHttpRedirectPolicyRule(httpsPort int32) *Rule {
...
+		newDg := InternalDataGroup{
+			Name:      name,
+			Partition: partition,
+		}
+		newDg.AddOrUpdateRecord(key, value)
+		intDgMap[mapKey] = &newDg
+	}
+}
+
+func (appMgr *Manager) updateRouteDataGroups(
```
f5gary: I just pushed a commit to address this.

pkg/appmanager/routing.go
```diff
@@ -343,7 +345,8 @@ func (appMgr *Manager) updatePassthroughRouteDataGroups(
 	return changed, nil
 }
 
-func updateDataGroupForRoute(
+// Update a data group map based on a passthrough a route object.
```
sjberman: nit: grammar "a passthrough a route object"

sjberman:
Feel free to squash and merge.

PR #291, opened 2017-08-24
Add node label selector

johscheuer:
Problem: If you use `nodeport` (which is the default) per default all nodes from the cluster will be added to the pool. In some cases you only want to use a subset of these nodes (e.q. only nodes with the label `node-role.kubernetes.io/node`) because not all nodes are reachable by the F5 or the nodes are already managed by F5 (but nor the Plugin). Currently we use to F5 to load balance traffic to the API servers (HA setup), this means that the Kubernetes master nodes are already present in F5 if you start the bigip-ctlr without a nodeLabelSelector the creation of the node pool for Kubernetes will fail since the bigip-ctlr can't create nodes that are already present.

### DIFF
--- a/pkg/pollers/nodePoller.go
+++ b/pkg/pollers/nodePoller.go
@@ -48,11 +48,13 @@ type nodePoller struct {
 	regListeners []PollListener
 	nodeCache    []v1.Node
 	lastError    error
+	nodeLabel    string
 }
 
 func NewNodePoller(
 	kubeClient kubernetes.Interface,
 	pollInterval time.Duration,
+	nodeLabel string,
 ) Poller {
 	np := &nodePoller{
 		kubeClient:   kubeClient,
@@ -61,6 +63,7 @@ func NewNodePoller(
 		addCh:        make(chan pollListener),
 		running:      false,
 		runningLock:  &sync.Mutex{},
+		nodeLabel:    nodeLabel,
 	}
 
 	log.Debugf("NodePoller object created: %p", np)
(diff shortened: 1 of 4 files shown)


sjsharma2001:
Hi Johannes, 
Thanks again for submitting the pull request. Sorry it took us some time to get the CLA stuff in place. Please look through the contents of [CONTRIBUTING.md](https://github.com/F5Networks/k8s-bigip-ctlr/blob/master/CONTRIBUTING.md), submit the CLA and we can start reviewing the PR.

johscheuer:
Thanks for the information I will sign the CLA and inform you.

johscheuer:
I signed and sent the CLA!

sjsharma2001:
Johannes, thanks for sending the CLA. We can start reviewing the PR now.

pkg/pollers/nodePoller.go
```diff
@@ -171,7 +174,9 @@ func (np *nodePoller) poller() {
 
 		if true == doPoll {
 			doPoll = false
-			nodes, err := np.kubeClient.Core().Nodes().List(metav1.ListOptions{})
+
+			// LabelSelector
+			nodes, err := np.kubeClient.Core().Nodes().List(metav1.ListOptions{LabelSelector: np.nodeLabel})
```
sjberman: If nodeLabel is an empty string, will this default to watching all nodes? Have you tested this?

sjberman: We found the test for it. Looks good.

sjberman:
Hi Johannes, just realized after the fact that the documentation for the new CLI option was missing. That's our fault for not bringing it up in the review, and we will add in the documentation for it. Please be mindful in any future code changes to include any documentation impacts. Thanks!

sjberman:
@johscheuer If I had this label on a node: "mylabel=test", would the correct formatting of the CLI argument be "--node-label-selector=mylabel=test"?

I'm running into an issue where Ingresses and Routes that reference services on a non-watched Node are still being configured, and I want to make sure that the CLI arg is configured properly.

sjberman:
Looks like that is the correct syntax, we have different issue. I'll open a ticket.

PR #296, opened 2017-08-25
Use iRule to support http redirect in 11.6.1

sjberman:
Problem: The action location parameter for our https redirect rule only
works in BIG-IP version 12.0.0+. Version 11.6.1 has a different format.

Solution: Use a generic iRule instead of a policy rule to set the http
redirect. This works in both BIG-IP versions.

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -647,6 +647,9 @@ func (appMgr *Manager) runImpl(stopCh <-chan struct{}) {
 	defer appMgr.vsQueue.ShutDown()
 	defer appMgr.nsQueue.ShutDown()
 
+	appMgr.addIRule(httpRedirectIRuleName, DEFAULT_PARTITION,
+		httpRedirectIRule(DEFAULT_HTTPS_PORT))
+
 	if nil != appMgr.routeClientV1 {
 		appMgr.addIRule(
 			sslPassthroughIRuleName, DEFAULT_PARTITION, sslPassthroughIRule())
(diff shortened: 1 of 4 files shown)


amudukutore:
Have you tested this on 12.X and 11.6.1? Is there a system test for this that was previously failing?

pkg/appmanager/routing.go
```diff
@@ -195,21 +195,13 @@ func processIngressRules(
...
-		Redirect:  true,
-		Request:   true,
-	}
-	rule := Rule{
-		Name:    httpRedirectRuleName,
-		Actions: []*action{&redirAction},
-		Ordinal: 0,
-	}
-	return &rule
+func httpRedirectIRule(port string) string {
```
f5gary: port should be passed in as int and %d used in the Sprintf() call rather than forcing the callers to all do the conversion.

pkg/appmanager/appManager.go
```diff
@@ -1206,11 +1209,15 @@ func (appMgr *Manager) handleIngressTls(
...
-		// State 2, apply HTTP redirect policy
-		log.Debugf("TLS: Applying HTTP redirect policy.")
-		policyName = fmt.Sprintf("%s-http-redirect",
-			rsCfg.Virtual.VirtualServerName)
-		rule = newHttpRedirectPolicyRule(httpsPort)
+		// State 2, set HTTP redirect iRule
+		log.Debugf("TLS: Applying HTTP redirect iRule.")
+		ruleName := fmt.Sprintf("/%s/%s", DEFAULT_PARTITION, httpRedirectIRuleName)
+		if httpsPort != DEFAULT_HTTPS_PORT {
+			ruleName = fmt.Sprintf("%s_%s", ruleName, httpsPort)
```
f5gary: Should be using %d for the httpsPort, not %s

sjberman:
@amudukutore Yes I've tested this on both versions. This is was found from a failing system test.

pkg/appmanager/routing.go
sjberman: Yep, good catch. I've gotten %s burned into my mind, I keep forgetting to change types.

PR #303, opened 2017-08-30
Convert unit tests to use Ginkgo/Gomega

sjberman:
Switch unit tests from using stretchr/testify package to use Ginkgo and Gomega.
This allows unit tests much more flexibility and readability. Ginkgo allows for BeforeEach
and AfterEach functions to reduce common code in unit tests. "Expect" keyword replaces assert and require for less ambiguity. Unit tests can be grouped based on context.


cmd/k8s-bigip-ctlr/main_test.go
```diff
@@ -17,19 +17,19 @@
 package main
 
 import (
-	"errors"
+	//"errors"
```
f5gary: Looks like this was left in by accident

PR #304, opened 2017-08-30
CCCL Integration

edarzins:
Moved all of the LTM config creation to the golang frontend, therefore
the python backend is just a pass-through to CCCL. Updated unit tests
too.

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -17,7 +17,6 @@
 package appmanager
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -1564,7 +1563,7 @@ func (appMgr *Manager) updatePoolMembersForNodePort(
 					svcKey, portSpec.NodePort)
 				rsCfg.MetaData.Active = true
 				rsCfg.MetaData.NodePort = portSpec.NodePort
-				rsCfg.Pools[index].PoolMemberAddrs =
+				rsCfg.Pools[index].Members =
 					appMgr.getEndpointsForNodePort(portSpec.NodePort)
 			}
 		}
@@ -1597,7 +1596,7 @@ func (appMgr *Manager) updatePoolMembersForCluster(
 			ipPorts := getEndpointsForService(portSpec.Name, eps)
 			log.Debugf("Found endpoints for backend %+v: %v", sKey, ipPorts)
 			rsCfg.MetaData.Active = true
-			rsCfg.Pools[index].PoolMemberAddrs = ipPorts
+			rsCfg.Pools[index].Members = ipPorts
 		}
 	}
 	return true, "", ""
@@ -1614,7 +1613,7 @@ func (appMgr *Manager) deactivateVirtualServer(
 	defer appMgr.resources.Unlock()
 	if rs, ok := appMgr.resources.Get(sKey, rsName); ok {
 		rsCfg.MetaData.Active = false
-		rsCfg.Pools[index].PoolMemberAddrs = nil
+		rsCfg.Pools[index].Members = nil
 		if !reflect.DeepEqual(rs, rsCfg) {
 			log.Debugf("Service delete matching backend %v %v deactivating config",
 				sKey, rsName)
@@ -1867,47 +1866,45 @@ func (appMgr *Manager) recordIngressEvent(ing *v1beta1.Ingress,
 func getEndpointsForService(
 	portName string,
 	eps *v1.Endpoints,
-) []string {
-	var ipPorts []string
+) []Member {
+	var members []Member
 
 	if eps == nil {
-		return ipPorts
+		return members
 	}
 
 	for _, subset := range eps.Subsets {
 		for _, p := range subset.Ports {
 			if portName == p.Name {
-				port := strconv.Itoa(int(p.Port))
 				for _, addr := range subset.Addresses {
-					var b bytes.Buffer
-					b.WriteString(addr.IP)
-					b.WriteRune(':')
-					b.WriteString(port)
-					ipPorts = append(ipPorts, b.String())
+					member := Member{
+						Address: addr.IP,
+						Port:    p.Port,
+						Session: "user-enabled",
+					}
+					members = append(members, member)
 				}
 			}
 		}
 	}
-	if 0 != len(ipPorts) {
-		sort.Strings(ipPorts)
-	}
-	return ipPorts
+	return members
 }
 
 func (appMgr *Manager) getEndpointsForNodePort(
 	nodePort int32,
-) []string {
-	port := strconv.Itoa(int(nodePort))
+) []Member {
 	nodes := appMgr.getNodesFromCache()
-	for i, v := range nodes {
-		var b bytes.Buffer
-		b.WriteString(v)
-		b.WriteRune(':')
-		b.WriteString(port)
-		nodes[i] = b.String()
+	var members []Member
+	for _, v := range nodes {
+		member := Member{
+			Address: v,
+			Port:    nodePort,
+			Session: "user-enabled",
+		}
+		members = append(members, member)
 	}
 
-	return nodes
+	return members
 }
 
 func handleConfigMapParseFailure(
@@ -1972,16 +1969,16 @@ func (appMgr *Manager) ProcessNodeUpdate(
 		if !reflect.DeepEqual(newNodes, appMgr.oldNodes) {
 			log.Infof("ProcessNodeUpdate: Change in Node state detected")
 			appMgr.resources.ForEach(func(key serviceKey, cfg *ResourceConfig) {
-				port := strconv.Itoa(int(cfg.MetaData.NodePort))
-				var newAddrPorts []string
+				var members []Member
 				for _, node := range newNodes {
-					var b bytes.Buffer
-					b.WriteString(node)
-					b.WriteRune(':')
-					b.WriteString(port)
-					newAddrPorts = append(newAddrPorts, b.String())
+					member := Member{
+						Address: node,
+						Port:    cfg.MetaData.NodePort,
+						Session: "user-enabled",
+					}
+					members = append(members, member)
 				}
-				cfg.Pools[0].PoolMemberAddrs = newAddrPorts
+				cfg.Pools[0].Members = members
 			})
 			// Output the Big-IP config
 			appMgr.outputConfigLocked()
--- a/pkg/appmanager/appManager_test.go
+++ b/pkg/appmanager/appManager_test.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"os"
 	"sort"
-	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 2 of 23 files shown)


pkg/appmanager/appManager.go
```diff
@@ -1867,47 +1866,45 @@ func (appMgr *Manager) recordIngressEvent(ing *v1beta1.Ingress,
...
+					members = append(members, member)
 				}
 			}
 		}
 	}
-	if 0 != len(ipPorts) {
-		sort.Strings(ipPorts)
-	}
-	return ipPorts
+	return members
```
f5gary: We aren't sorting anymore - is this something we are counting on CCCL to do for us?

pkg/appmanager/types.go
```diff
@@ -57,52 +61,79 @@ type (
...
-		Partition       string   `json:"partition"`
+		Partition       string   `json:"partition,omitempty"`
 		Balance         string   `json:"loadBalancingMode"`
-		ServiceName     string   `json:"serviceName"`
-		ServicePort     int32    `json:"servicePort"`
-		PoolMemberAddrs []string `json:"poolMemberAddrs"`
-		MonitorNames    []string `json:"monitor"`
+		ServiceName     string   `json:"serviceName,omitempty"`
+		ServicePort     int32    `json:"servicePort,omitempty"`
+		Members			[]Member `json:"members"`
```
f5gary: This formatting looks messed up. Is "make verify" passing?

pkg/appmanager/outputConfig.go
```diff
@@ -61,26 +119,105 @@ func (appMgr *Manager) outputConfigLocked() {
...
+			} else if mode == "tcp" {
+				resources[partition].Virtuals[i].IpProtocol = "tcp"
+				profile := ProfileRef{Partition: "Common", Name: "tcp", Context: "all"}
+				resources[partition].Virtuals[i].Profiles = append(resources[partition].Virtuals[i].Profiles, profile)
+			} else if mode == "udp" {
+				resources[partition].Virtuals[i].IpProtocol = "udp"
+			}
+
+			// Parse the SSL profile into partition and name
+			if nil != resources[partition].Virtuals[i].SslProfile {
```
f5gary: There is already a function that makes it easier to deal with Virtual.SslProfile. Calling resources[partition].Virtuals[i].GetFrontendSslProfileNames() will return an array that you can just loop through here.

pkg/appmanager/appManager.go
edarzins: It doesn't matter for CCCL, was there a reason for sorting them?

pkg/appmanager/outputConfig.go
edarzins: OK!

pkg/appmanager/types.go
edarzins: It's not passing right now, I will make sure it's clean before the next push.

pkg/appmanager/types.go
```diff
@@ -57,52 +61,79 @@ type (
...
+		Partition string `json:"partition,omitempty"`
 
 		// VirtualServer parameters
-		Balance        string          `json:"balance,omitempty"`
-		Mode           string          `json:"mode,omitempty"`
-		VirtualAddress *virtualAddress `json:"virtualAddress,omitempty"`
-		SslProfile     *sslProfile     `json:"sslProfile,omitempty"`
-		Policies       []nameRef       `json:"policies,omitempty"`
-		IRules         []string        `json:"rules,omitempty"`
-		// FIXME: All profiles should reside in Profiles, just server ssl ones now.
```
f5gary: This comment shouldn't be removed since SslProfile is still here (we have profiles in more than one spot still).

pkg/appmanager/resourceConfig.go
```diff
@@ -577,15 +577,19 @@ func copyConfigMap(cfg *ResourceConfig, cfgMap *ConfigMap) {
 			name = fmt.Sprintf("%s", cfg.Virtual.VirtualServerName)
 		}
 		monitor := Monitor{
-			Name:      name,
+			// Append the protocol to the monitor names to differentiate them.
+			// Also add a monitor index to the name to be consistent with the
+			// marathon-bigip-ctlr. Since the monitor names are already unique here,
+			// appending a '0' is sufficient.
+			Name:      name + "_0_" + mon.Protocol,
```
f5gary: I understand the need to add the protocol to the end here, but it seems like the "0" shouldn't be added here; instead the "if index > 0" block above changed to just format the name as expected. As this is now, the first monitor might look like "name_0_http", but the second would look like "name_1_0_http" which seems wrong ("name_1_http" seems better).

pkg/appmanager/outputConfig.go
```diff
@@ -61,26 +119,105 @@ func (appMgr *Manager) outputConfigLocked() {
...
+				// 'f5ProfileName' or 'f5ProfileNames', not both.
+				if "" != resources[partition].Virtuals[i].SslProfile.F5ProfileName {
+					resources[partition].Virtuals[i].Profiles =
+						appendSslProfile(resources[partition].Virtuals[i].Profiles,
+										 resources[partition].Virtuals[i].SslProfile.F5ProfileName,
+										 "clientside")
+                         } else {
+					for _, p := range resources[partition].Virtuals[i].SslProfile.F5ProfileNames {
+						resources[partition].Virtuals[i].Profiles =
+							appendSslProfile(resources[partition].Virtuals[i].Profiles, p, "clientside")
```
f5gary: We have a constant defined for "clientside", so please use customProfileClient instead.

pkg/appmanager/appManager.go
f5gary: I'm not sure why it was done here before (I agree it doesn't matter to CCCL) but I saw it was removed. We aren't doing it in the NodePort case, so I think it's fine to remove the sorting here.

edarzins: Removing the sorting was intentional, it does not matter for CCCL.

pkg/appmanager/resourceConfig.go
edarzins: Agree

pkg/appmanager/types.go
edarzins: sslProfile just contains the profile names parsed from the ConfigMap. I convert each of these to a ProfileRef and put them in Profiles, which gets shipped to CCCL.

f5gary: Yes, but our goal is to get the go code to eventually be closer to the CCCL layout and not the configmap + CCCL layouts combined, so in that scenario all profiles would go into Profiles and all the quirky configmap stuff would be translated when configmaps are parsed. So the comment is still valid for the next step.

edarzins: OK, I will keep the comment.

pkg/appmanager/appManager_test.go
```diff
@@ -267,23 +266,23 @@ var configmapIApp2 string = string(`{
...
...ngMode":"round-robin","members":[{"address":"127.0.0.1","port":37001,"session":"user-enabled"},{"address":"127.0.0.2","port":37001,"session":"user-enabled"},{"address":"127.0.0.3","port":37001,"session":"user-enabled"}],"monitors":null},{"name":"default_foomap","loadBalancingMode":"round-robin","members":[{"address":"127.0.0.1","port":30001,"session":"user-enabled"},{"address":"127.0.0.2","port":30001,"session":"user-enabled"},{"address":"127.0.0.3","port":30001,"session":"user-enabled"}],"monitors":["/velcro/default_foomap_0_tcp"]},{"name":"default_foomap8080","loadBalancingMode":"round-robin","members":[{"address":"127.0.0.1","port":38001,"session":"user-enabled"},{"address":"127.0.0.2","port":38001,"session":"user-enabled"},{"address":"127.0.0.3","port":38001,"session":"user-enabled"}],"monitors":null},{"name":"default_foomap9090","loadBalancingMode":"round-robin","members":[{"address":"127.0.0.1","port":39001,"session":"user-enabled"},{"address":"127.0.0.2","port":39001,"session":"user-enabled"},{"address":"127.0.0.3","port":39001,"session":"user-enabled"}],"monitors":null}],"monitors":[{"name":"default_foomap_0_tcp","interval":30,"type":"tcp","send":"GET /","timeout":20}]}}}`)
```
f5gary: The unit tests in this file were recently converted to another test framework, and as a result I think you have a fair amount of work to resolve those conflicts.

f5gary:
How do the system tests look? I think this looks good (besides the unit test conflicts), though we will need some followup work in the controller to get the internal data representation closer to the CCCL format.

pkg/appmanager/appManager_test.go
edarzins: I rebased and resolved all merge conflict yesterday.

edarzins:
k8s and OpenShift systests are 100% PASS for BIG-IP v12.1.0. A few failures for v11.6.1, but the same that are failing for master.

pkg/appmanager/appManager_test.go
f5gary: Oh ok. Github was telling me it was conflicting but it isn't anymore.

edarzins:
I will squash the commits, retest image, and then do a new push.

pkg/appmanager/outputConfig.go
```diff
@@ -61,26 +119,95 @@ func (appMgr *Manager) outputConfigLocked() {
...
 	for _, intDg := range appMgr.intDgMap {
-		resources.InternalDataGroups = append(resources.InternalDataGroups, *intDg)
+		initPartitionData(resources, intDg.Partition)
+		resources[intDg.Partition].InternalDataGroups = append(resources[intDg.Partition].InternalDataGroups, *intDg)
+	}
+
+	// Update resources to conform to the CCCL schema and empty out unneeded fields
+	// so they will be stripped out by the JSON marshaller
+	for partition, partitionConfig := range resources {
+		for i, _ := range partitionConfig.Virtuals {
```
f5yacobucci: This looks like a good candidate for a sync.WorkGroup and goroutine per for loop.
```
var wg sync.WorkGroup
wg.Add(1)
go runForLoop {
  ...
  wg.Done()
}
wg.Add(1)
go runForLoop {
  ...
  wg.Done()
}
wg.Wait()
```

edarzins: Are you thinking a goroutine per partition or per resource slice (i.e. Virtuals, Pools, etc)?

f5yacobucci: I was thinking more for resource slice but if all these loop iterations work on independent chunks of data then you could go further and make a goroutine per partition that then also creates goroutines to independently work on each of their sections.

python/bigipconfigdriver.py
```diff
@@ -366,10 +353,10 @@ def create_config_kubernetes(partition, config):
     f5 = {'ltm': {}, 'network': {}}
     if 'openshift-sdn' in config:
         f5['network'] = create_network_config_kubernetes(config)
-    if 'resources' in config and 'virtualServers' in config['resources']:
-        f5['ltm'] = create_ltm_config_kubernetes(partition,
-                                                 config['resources'])
+    if 'resources' in config and partition in config['resources']:
```
f5yacobucci: What this ends up doing is stamping 'network' for every partition running extra apply fdb configs. What would be better here is to keep ltm and network separate.

In do reset we can just loop over managers and get the partition config. Then after that loop do one single network call.

pkg/appmanager/outputConfig.go
edarzins: Agree

python/bigipconfigdriver.py
edarzins: Agree, good catch!

amudukutore:
@edarzins - Please close out internal issue k8s-bigip-ctlr/issues/203 once this is in. Thanks!

PR #309, opened 2017-09-06
Fix SNIDefault for Routes

sjberman:
Problem: SNIDefault was not being set for custom profiles when configuring
https Routes. The server name was set, but a default SNI was not.

Solution: Set the default SNI for the first custom profile for a virtual server.
All subsequent profiles for that same virtual server will not be set for default SNI.

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -161,7 +161,9 @@ func NewManager(params *Params) *Manager {
 }
 
 func (appMgr *Manager) loadDefaultCert(
-	namespace string,
+	namespace,
+	serverName,
+	vsName string,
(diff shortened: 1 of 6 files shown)


pkg/appmanager/appManager.go
```diff
@@ -1152,6 +1152,11 @@ func (appMgr *Manager) setClientSslProfile(
 			Cert:       route.Spec.TLS.Certificate,
 			Key:        route.Spec.TLS.Key,
 			ServerName: route.Spec.Host,
+			VSName:     rsCfg.Virtual.VirtualServerName,
+		}
```
sjberman: Use better NewCustomProfile factory to build this.

PR #312, opened 2017-09-07
Fix SIGSEV on non-"f5" valued class annotation

### LINKED ISSUE #311: Using kubernetes.io/ingress.class other than "f5" causes panic (SIGSEGV)
Problem: If the k8s-bigip-ctlr is running, and an Ingress is configured with the "kubernetes.io/ingress.class" annotation, and that annotation has a value other than "f5", the k8s-bigip-ctlr will crash:

    /build/src/github.com/F5Networks/k8s-bigip-ctlr/vendor/k8s.io/client-go/tools/cache/shared_informer.go:547
    /usr/local/go/src/runtime/asm_amd64.s:2086
    panic: runtime error: invalid memory address or nil pointer dereference [recovered]
            panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x4c3372]

Workaround: The kubernes.io/ingress.class annotation is only required if you need to run multiple different ingress controllers for the same kubernetes cluster (for instance, a k8s-bigip-ctlr and an nginx ingress controller).  If you do not need to do this, then do not specify a kubernetes.io/ingress.class annotation until this bug is fixed.

sjberman:
Problem: When using a kubernetes ingress class other than "f5", the controller
hit a segfault, due to trying to access a field in a nil struct.

Solution: Change the way the required value is grabbed, and use outputConfigLocked instead of
outputConfig, since the calling function is already locked.

Fixes #311

### DIFF
--- a/pkg/appmanager/appManager_test.go
+++ b/pkg/appmanager/appManager_test.go
@@ -2464,6 +2464,24 @@ var _ = Describe("AppManager Tests", func() {
 				Expect(r).To(BeTrue(), "Ingress resource should be processed.")
 				Expect(resources.Count()).To(Equal(0))
 
+				// Shouldn't process Ingress with non-F5 class
+				// https://github.com/F5Networks/k8s-bigip-ctlr/issues/311
+				ingressNotf5 := test.NewIngress("ingress-bad", "1", namespace, ingressConfig,
+					map[string]string{
+						"kubernetes.io/ingress.class": "notf5",
+					})
+				r = mockMgr.addIngress(ingressNotf5)
+				Expect(r).To(BeFalse(), "Ingress resource should not be processed.")
+				Expect(resources.Count()).To(Equal(0))
+				ingressNotf5.Annotations["kubernetes.io/ingress.class"] = "f5"
+				r = mockMgr.updateIngress(ingressNotf5)
+				Expect(r).To(BeTrue(), "Ingress resource should be processed when flipping from notf5 to f5.")
+				Expect(resources.Count()).To(Equal(1))
+				ingressNotf5.Annotations["kubernetes.io/ingress.class"] = "notf5again"
+				r = mockMgr.updateIngress(ingressNotf5)
+				Expect(r).To(BeFalse(), "Ingress resource should be destroyed when flipping from f5 to notf5again.")
+				Expect(resources.Count()).To(Equal(0))
+
 				// Multi-service Ingress
 				ingressConfig = v1beta1.IngressSpec{
 					Rules: []v1beta1.IngressRule{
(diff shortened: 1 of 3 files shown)


sjberman:
This will need to be cherry picked to 1.1-stable

pkg/appmanager/appManager_test.go
```diff
@@ -2464,6 +2464,15 @@ var _ = Describe("AppManager Tests", func() {
 				Expect(r).To(BeTrue(), "Ingress resource should be processed.")
 				Expect(resources.Count()).To(Equal(0))
 
+				// Shouldn't process Ingress with non-F5 class
+				ingressNotf5 := test.NewIngress("ingress-bad", "1", namespace, ingressConfig,
```
andrewjjenkins: Could you add a comment here pointing back to the bug?

// https://github.com/F5Networks/k8s-bigip-ctlr/issues/311

pkg/appmanager/validateResources.go
```diff
@@ -123,7 +123,7 @@ func (appMgr *Manager) checkValidIngress(
 					appMgr.resources.Delete(key, rsName)
 				}
 			}
-			appMgr.outputConfig()
+			appMgr.outputConfigLocked()
```
andrewjjenkins: As we chatted about, maybe it'd be better to only output the config if you had to delete a resource from the appMgr (to avoid writing out a new config if an Ingress we don't care about is created)

pkg/appmanager/appManager_test.go
```diff
@@ -2464,6 +2464,15 @@ var _ = Describe("AppManager Tests", func() {
...
+					map[string]string{
+						"kubernetes.io/ingress.class": "notf5",
+					})
+				r = mockMgr.addIngress(ingressNotf5)
+				Expect(r).To(BeFalse(), "Ingress resource should not be processed.")
+				Expect(resources.Count()).To(Equal(0))
+
```
andrewjjenkins: What are your thoughts on also testing the "modify existing ingress.class annotation" case?

I added this code here and it worked (and I'd sleep better at night if we had some test case like this):

```
ingressNotf5.Annotations["kubernetes.io/ingress.class"] = "f5"
r = mockMgr.updateIngress(ingressNotf5)
Expect(r).To(BeTrue(), "Ingress resource should be processed when flipping from notf5 to f5")
Expect(resources.Count()).To(Equal(1))
ingressNotf5.Annotations["kubernetes.io/ingress.class"] = "notf5again"
r = mockMgr.updateIngress(ingressNotf5)
Expect(r).To(BeFalse(), "Ingress resource should be destroyed when flipping from f5 to notf5again")
Expect(resources.Count()).To(Equal(0))
```

PR #316, opened 2017-09-08
Node label selector CLI docs

sjberman:
Problem: Missing documentation for the new CLI option that allows for specifying a node label
for the controller to watch for.

Solution: Added the CLI description to the README.

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -64,85 +64,93 @@ Controller Configuration Parameters
 -----------------------------------
 The configuration parameters below are global to the |kctlr|.
 
-+--------------------+---------+----------+-------------+-----------------------------------------+----------------+
-| Parameter          | Type    | Required | Default     | Description                             | Allowed Values |
-+====================+=========+==========+=============+=========================================+================+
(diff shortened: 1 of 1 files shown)


amudukutore:
@sjberman - are there going to be some changes to the solution docs to explain when and why a user may want to use this option?

docs/README.rst
```diff
@@ -59,86 +59,94 @@ Controller Configuration Parameters
...
+|                     |         |          |             |                                         |                |
+|                     |         |          |             | Only applicable in the OpenShift        |                |
+|                     |         |          |             | environment.                            |                |
++---------------------+---------+----------+-------------+-----------------------------------------+----------------+
 
+.. note::
+
+  Use the ``node-label-selector`` parameter if you only want the controller to manage specific nodes from the cluster.
+  For example, the BIG-IP device may not be able to reach certain nodes, or the BIG-IP device already manages certain
+  nodes. Therefore, the controller should only watch the nodes outside of these contexts (by using a label).
```
f5yacobucci: I think this might be a little clearer for the last sentence:

Therefore, the controller should only watch the nodes that match the environmental constraints (by using a label).

PR #323, opened 2017-09-12
Controller should support Node Taints for k8s 1.6

### LINKED ISSUE #321: Controller should support Node Taints for k8s 1.6
The controller sometimes wishes to identify Unschedulable nodes in order to prevent sending traffic to them - traffic that is guaranteed a second hop.

In 1.6 Taints are beta and fully supported in 1.7. Nodes may no longer be marked Unschedulable but be tainted as NoSchedule. Creating the same behavior but failing the controller test of watching the Unschedulable field.

f5yacobucci:
Problem:
The controller sometimes wishes to identify Unschedulable nodes in order
to prevent sending traffic to them - traffic that is guaranteed a second
hop.

In 1.6 Taints are beta and fully supported in 1.7. Nodes may no longer
be marked Unschedulable but be tainted as NoSchedule. Creating the same
behavior but failing the controller test of watching the Unschedulable
field.

Solution:
Update controller to look for and prefer the Node.Spec.Taint setting and
then defer to the Node.Spec.Unschedulable setting - if a node selector
label has been configured appManager will ignore taint.

Tests: new unit tests, and test_restore.py / test_e2e.py

Fixes #321

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -97,19 +97,22 @@ type Manager struct {
 	eventSource   v1.EventSource
 	// Route configurations
 	routeConfig RouteConfig
+	// Currently configured node label selector
+	nodeLabelSelector string
 }
 
 // Struct to allow NewManager to receive all or only specific parameters.
 type Params struct {
-	KubeClient      kubernetes.Interface
-	restClient      rest.Interface // package local for unit testing only
-	RouteClientV1   rest.Interface
-	ConfigWriter    writer.Writer
-	UseNodeInternal bool
-	IsNodePort      bool
-	RouteConfig     RouteConfig
-	InitialState    bool                 // Unit testing only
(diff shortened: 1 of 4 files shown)


pkg/appmanager/appManager.go
```diff
@@ -96,20 +96,22 @@ type Manager struct {
 	eventRecorder record.EventRecorder
 	eventSource   v1.EventSource
 	// Route configurations
-	routeConfig RouteConfig
+	routeConfig       RouteConfig
+	nodeLabelSelector string
```
sjberman: I would add a newline or comment between these two fields (just per the convention of the struct).

f5yacobucci:
Yes, this works in 1.4 and in 1.6. This is something that we want to include in 1.2 to keep behavior the same if node taints are used.

pkg/appmanager/appManager.go
f5yacobucci: +1

PR #333, opened 2017-09-18
Support Routes in multiple namespaces

### LINKED ISSUE #331: Can't create virtuals for Routes in multiple namespaces
The current naming convention for virtual servers for Routes includes the namespace. This causes the controller to attempt to create multiple sets of virtual servers for different namespaces. The problem is that the controller is only given one IP address ('route-vserver-addr'), and therefore there is a conflict when creating the virtual servers.

The fix for this is to simply use one set of http and https virtual servers for all Routes in all namespaces, rather than separating by namespace.

sjberman:
Problem: Routes in multiple namespaces were not supported, since a single IP address is supplied
to the controller, and therefore would cause conflicts when attempting to create a new set of virtuals
for a new namespace.

Solution: Rename the virtuals for OpenShift Routes, and only create 2 vservers for all routes in all namespaces.

Fixes #331

### DIFF
--- a/pkg/appmanager/appManager_test.go
+++ b/pkg/appmanager/appManager_test.go
@@ -2861,244 +2861,253 @@ var _ = Describe("AppManager Tests", func() {
 				Expect(len(customProfiles)).To(Equal(1))
 			})
 
-			It("configures virtual servers via Routes", func() {
-				spec := routeapi.RouteSpec{
-					Host: "foobar.com",
-					Path: "/foo",
-					To: routeapi.RouteTargetReference{
-						Kind: "Service",
-						Name: "foo",
-					},
-					TLS: &routeapi.TLSConfig{
-						Termination: "edge",
-						Certificate: "cert",
-						Key:         "key",
-					},
-				}
-				route := test.NewRoute("route", "1", namespace, spec)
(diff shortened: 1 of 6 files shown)


pkg/appmanager/appManager_test.go
```diff
@@ -2888,7 +2888,7 @@ var _ = Describe("AppManager Tests", func() {
 				Expect(resources.Count()).To(Equal(2))
 
 				rs, ok := resources.Get(
-					serviceKey{"foo", 80, "default"}, "openshift_default_https")
+					serviceKey{"foo", 80, "default"}, "https-ose-vserver")
```
lihongan: Can user specify virtual server names other than "https-ose-vserver" and "ose-vserver" ? The name can be configured when start OpenShift F5 router.  
```
$ oadm router \
    --type=f5-router \
    --external-host-http-vserver=ose-vserver \
    --external-host-https-vserver=https-ose-vserver \
```
see https://docs.openshift.com/container-platform/3.5/install_config/router/f5_router.html

sjberman: Sure, I can add in the ability to override these default names via command line arguments.

PR #334, opened 2017-09-19
Create default client and server profiles for SNI

sjberman:
Problem: The controller was using the first https route's cert/key
as the default for SNI for the https virtual server. There was a bug
that either caused none of the Routes to be used, or more than one.

Upon further discussion, we concluded that we don't want to use the Routes'
certs/keys as the default SNI profile, and instead a generic or user-specified default.

Solution: Route certs/keys will no longer be used as default SNI profiles for the https
virtual server. Instead, we will allow the user to create client and server ssl profiles and
specify them via CLI. If none are specified, the controller creates a generic profile, inherited
from either the clientssl or serverssl profiles, and use those as default for SNI.

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -121,6 +121,8 @@ type RouteConfig struct {
 	RouteLabel  string
 	HttpVs      string
 	HttpsVs     string
+	ClientSSL   string
+	ServerSSL   string
 }
 
 // Create and return a new app manager that meets the Manager interface
@@ -168,8 +170,7 @@ func NewManager(params *Params) *Manager {
 
 func (appMgr *Manager) loadDefaultCert(
 	namespace,
-	serverName,
-	vsName string,
+	serverName string,
 ) (*ProfileRef, bool) {
 	// OpenShift will put the default server SSL cert on each pod. We create a
 	// server SSL profile for it and associate it to any reencrypt routes that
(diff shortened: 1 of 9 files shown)


pkg/appmanager/appManager.go
```diff
@@ -1199,6 +1221,31 @@ func (appMgr *Manager) setServerSslProfile(
...
+		rsCfg.Virtual.AddOrUpdateProfile(profile)
+	} else {
+		// No provided name, so we create a default
+		profile := ProfileRef{
+			Name:      "default-route-serverssl",
+			Partition: rsCfg.Virtual.Partition,
+			Context:   customProfileServer,
+		}
+		// This is just a basic profile, so we don't need all the fields
+		cp := NewCustomProfile(profile, "", "", "", true, appMgr.customProfiles)
```
f5yacobucci: If the user doesn't configure a client or server ssl default are we creating a new default profile every time we set the profiles - even if we've already created one. Shouldn't this happen for just the first set? Then we can check the profs each time through?

pkg/appmanager/appManager.go
```diff
@@ -1199,6 +1221,31 @@ func (appMgr *Manager) setServerSslProfile(
 	rsCfg *ResourceConfig,
 	route *routeapi.Route,
 ) {
+	// First handle the Default for SNI profile
+	if appMgr.routeConfig.ServerSsl != "" {
+		// User has provided a name
+		profile := ProfileRef{
+			Name:      appMgr.routeConfig.ServerSsl,
```
f5yacobucci: Idiomatic go prefers acronyms with all caps, ServerSSL, this name may cause extra linter warnings.

pkg/appmanager/appManager.go
sjberman: This does create a new profile every time, but it is always the same, and overrides the previous entry that stored the same profile. Nothing different is written out, but we are calling this piece of code every time.

A similar process happens later in this function when processing the Route profile. I can add a check here to only create the new profile if it doesn't already exist.

f5yacobucci: Spoke offline, can do exists check from map lookup.

Issue #337, opened 2017-09-20
Exclude release notes from docs linkcheck


amudukutore:
Release notes may contain references to future links in the solution documentation that are not available until release time.

PR #338, opened 2017-09-20
Exclude release notes from docks link checker

amudukutore:
Problem:
linkcheck fails due to links in release notes not
yet live.

Solution:
Remove RELEASE-NOTES.rst file prior to linkcheck. The html file has
already been generated at this point and the source is no longer needed.

Testing:
Tested that the following command passed linkcheck with changes (these
failed previously):
./build-tools/docker-docs.sh build-tools/make-docs.sh

affects-branches: master

Fixes: #337

### DIFF
--- a/build-tools/make-docs.sh
+++ b/build-tools/make-docs.sh
@@ -6,6 +6,12 @@ set -e
 echo "Building docs and checking links with Sphinx"
 rm -rf docs/_build
 make -C docs html
+
+# Exclude release notes from linkcheck when running in travis.
+# Note: we've already rendered these - its safe to remove the unneeded source files.
+if [ "$TRAVIS" = true ]; then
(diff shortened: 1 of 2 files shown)


build-tools/make-docs.sh
```diff
@@ -6,6 +6,9 @@ set -e
 echo "Building docs and checking links with Sphinx"
 rm -rf docs/_build
 make -C docs html
+# Exclude release notes from linkcheck.
+# Note: we've already rendered these - its safe to remove the unneeded source files.
+rm -f docs/RELEASE-NOTES.rst
```
jputrino: If you're converting to .MD, then that change needs to be reflected here.

amudukutore:
I've added the check to do the exclude of the file only for travis build per our offline chat.

PR #345, opened 2017-09-22
Add redirect limitation to Release Notes

sjberman:
Mentioned the limitation that the controller currently cannot specify individual
redirect policies per route. Redirect applies to all routes on the virtual

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -20,6 +20,13 @@ Bug Fixes
 * Properly configure http redirect rules on v11.6.1 BIG-IP systems.
 * Failed configurations for objects do not prevent future configurations from happening.
 
+Limitations
+~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


docs/RELEASE-NOTES.rst
```diff
@@ -20,6 +20,12 @@ Bug Fixes
 * Properly configure http redirect rules on v11.6.1 BIG-IP systems.
 * Failed configurations for objects do not prevent future configurations from happening.
 
+Limitations
+~~~~~~~~~~~
+
+* OpenShift - Does not currently support redirect for individual Routes. If a Route specifies 
+  "insecureEdgeTerminationPolicy" as "Redirect", the http virtual server will enable this policy for all Routes.
```
amudukutore: Can we add a link to the open issue at the end of this (much like we do for bug fixes)? Looks good aside from that.

sjberman: Added

PR #349, opened 2017-09-26
Update version info

amudukutore:
Problem:
Version information needs to be updated so that we have proper
version information in releast notes, and also publish the docs
to the proper location.

Solution:
Updated the version information to v1.2.


jputrino:
Ok to merge once tests pass.

PR #354, opened 2017-09-26
IPAM: DNS resolution

sjberman:
Add support to resolve Ingress hosts into virtual server IP addresses. There is a new command line argument, "--resolve-ingress-names" which enables this feature. When specified with the string "LOOKUP", the controller will use local DNS to resolve the first hostname in an Ingress resource, and use that IP address as the virtual server address. A user can also specify a custom DNS hostname or IP address in the CLI, and the controller will instead use that DNS to resolve the Ingress host name. The controller populates the 'virtual-server.f5.com/ip' annotation with the resolved IP address.

If an Ingress doesn't have a host name, or the first host name is empty, no resolution occurs. If resolution fails for a hostname, there are no retries. If the 'virtual-server.f5.com/ip' annotation is already set, then no resolution is attempted, even if the CLI argument is supplied.


cmd/k8s-bigip-ctlr/main.go
```diff
@@ -158,9 +171,16 @@ func _init() {
 		"Optional, specify whether or not to manage Route resources")
 	nodeLabelSelector = kubeFlags.String("node-label-selector", "",
 		"Optional, used to watch only for nodes with this label")
+	resolveIngNames = kubeFlags.String("resolve-ingress-names", "",
+		"Optional, direct the controller to resolve host names in Ingresses into IP addresses. "+
+			"'LOOKUP' option will use the controller's built-in DNS. "+
```
f5gary: **The** 'LOOKUP' option...

pkg/appmanager/appManager.go
```diff
@@ -1978,6 +1990,95 @@ func (appMgr *Manager) recordIngressEvent(ing *v1beta1.Ingress,
...
+				logDNSError(fmt.Sprintf("Error while resolving host '%s': %s",
+					appMgr.resolveIng, err))
+				return
+			}
+			customDNS = netIPs[0].String()
+		}
+		client := dns.Client{}
+		msg := dns.Msg{}
+		msg.SetQuestion(host+".", dns.TypeA)
+		res, _, err := client.Exchange(&msg, customDNS+":53")
```
f5gary: It is possible that the custom DNS server may be running on a different port than 53, so we should handle that case by only appending ':53' in the case where customDNS does not already contain a port.

PR #355, opened 2017-09-27
Improve travis build times

sjberman:
Problem: The travis build was taking a long time due to the devel containers
being built more times than necessary.

Solution: Remove unnecessary devel builds. Also specified the number of compilers
used in the Ginkgo build. This *may* help with ginkgo run times, though
that has yet to be seen.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,6 @@ script:
   - export CLEAN_BUILD=true
   - export BASE_OS=alpine
   - ./build-tools/build-devel-image.sh
-  - ./build-tools/run-in-docker.sh make verify
   - ./build-tools/build-debug-artifacts.sh
   - ./build-tools/build-release-artifacts.sh
   - ./build-tools/build-release-images.sh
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -44,10 +43,10 @@ deploy:
   - provider: script
     skip_cleanup: true
     on:
-      branch: 1.1-stable
+      branch: 1.2-stable
```
amudukutore: I'd leave this change out - this will be addressed separately to make the docs deploy section version agnostic.

.travis.yml
```diff
@@ -44,10 +43,10 @@ deploy:
...
       repo: F5Networks/k8s-bigip-ctlr
     script:
-      - ./build-tools/deploy-docs.sh publish-product-docs-to-prod connectors/k8s-bigip-ctlr v1.1
+      - ./build-tools/deploy-docs.sh publish-product-docs-to-prod connectors/k8s-bigip-ctlr v1.2
```
amudukutore: Same comment as above.

amudukutore:
I know build times are variable but it would be nice to see (just a comment here will suffice) how much they were on average prior to this change and what it is with these changes in place.

sjberman:
Current build time now is about 15 to 16 minutes, increasing build times by at least 10 minutes. RHEL builds will also likely see a significant increase.

amudukutore:
@sjberman - can you please cherry-pick these into 1.2-stable? Thanks.

PR #362, opened 2017-09-29
Fix coverage badge reference

michaeldayreads:
Problem:
Coveralls badge pointed at "master" is showing coverage is "unavilaable".

Solution:
Use `?branch=HEAD` to indicate which number to post.


michaeldayreads:
@dramich functioning badge can be confirmed on my branch:
https://github.com/michaeldayreads/k8s-bigip-ctlr/blob/badge-fix/README.md

PR #366, opened 2017-10-02
Add docker latest tag and docker label to allow pinned image for tests

michaeldayreads:
Problem:
- Dockerhub images did not have "latest" tag
- images need to be able to be specifically referenced in automated regression tests

Solution:
- Updated tagging to included latest
- Added timestamp to build tag and included that information as a docker label to make it available to docker inspect in regression set up.


michaeldayreads:
https://hub.docker.com/r/michaeldayreads/k8s-bigip-ctlr/tags/

michaeldayreads:
@amudukutore - updated

PR #373, opened 2017-10-04
Add BUILD_STAMP to rhel image

michaeldayreads:
Problem:
rhel image build process did not export BUILD_STAMP for regression tests to pin image.

Solution:
Added BUILD_STAMP to rhel build using same labeling conventions for other rhel tags.


Jenkinsfile
```diff
@@ -30,6 +31,7 @@ pipeline {
           docker push "$IMG_TAG"
           docker push "$BASE_PUSH_TARGET:devel-$GIT_BRANCH-$BASE_OS"
           docker push "$BASE_PUSH_TARGET:devel-$GIT_BRANCH-n-$BUILD_NUMBER-id-$BUILD_ID-$BASE_OS"
```
amudukutore: Do we need this tag pushed as well or is it sufficient to just push the new one with the added timestamp?

michaeldayreads:
@amudukutore updated per offline conversation.

amudukutore:
Thanks for making these changes.

PR #374, opened 2017-10-04
Product doc fixes

### LINKED ISSUE #367: Ingress partition annotation is optional, not required (docs)
The README doc states that the 'virtual-server.f5.com/partition' annotation is required for Ingress resources, but it is optional.

If the annotation isn't provided, we use the partition that was passed to the controller via CLI.

sjberman:
Problem:
  1. Ingress partition annotation was listed as required, but is optional.
  2. OpenShift Route port information was not quite clear.

Solution:
  1. Listed the partition annotation as optional.
  2. Cleared up some ambiguity with the Route port information.

Fixes #367

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -429,7 +429,7 @@ Supported annotations
 +====================================+=============+===========+=====================================================================================+=============+
 | virtual-server.f5.com/ip           | string      | Required  | The IP address you want to assign to the virtual server.                            | N/A         |
 +------------------------------------+-------------+-----------+-------------------------------------------------------------------------------------+-------------+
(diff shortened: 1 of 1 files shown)


docs/README.rst
```diff
@@ -541,7 +541,8 @@ Supported Route Configurations
 
    - For Edge and Unsecured Route types, the default backend port is 80.
 
-   - To expose a Service on any other ports. **specify the port in the Route config's "Port" field**.
+   - To expose a Service on any other port, **specify the port in the Route config's "Port: TargetPort" field**.
```
jputrino: Change to "specify the port number..." and remove the nested bullet point.

sjberman: Fixed

PR #376, opened 2017-10-04
Allow users to disable Secret SSL loading

sjberman:
Problem: Users may be uncomfortable with the controller looking for Secrets for custom SSL
profile loading. Even if a user doesn't provide a secret name to the Ingress or ConfigMap,
the controller still looks for a secret before defaulting to a BIG-IP profile.

Solution: Add a CLI flag to disable the controller from ever trying to look for Secrets to
load SSL profiles from. By default, this flag is enabled to true (backwards compatability),
but the user can set to false if desired.


sjberman:
Passive voice issue has been fixed.

PR #382, opened 2017-10-10
OpenShift Route health monitor support

sjberman:
Problem: Users would like to be able to configure health monitors for their Route resources.

Solution: Use the existing health annotation for Ingress and adapt it for use with Route resources.
The annotation remains the exact same, and the processing for Routes works similar to single service
Ingress.


pkg/appmanager/resourceConfig.go
```diff
@@ -931,14 +931,56 @@ func (rc *ResourceConfig) SetMonitor(pool *Pool, monitor Monitor) {
...
+	return updated
+}
+
+func (rc *ResourceConfig) RemoveMonitor(pool, monitor string) bool {
+	var removed bool
+	for i, pl := range rc.Pools {
+		if pl.Name == pool {
+			for j, mon := range pl.MonitorNames {
+				if mon == monitor {
+					if i >= len(pl.MonitorNames)-1 {
```
f5gary: I don't understand the use of 'i' in this context, and why you wouldn't always do the 'else' case here.

sjberman: That should be a 'j'. If we are looking at the last monitor in the list, we can't access j+1 in the 'else' case.

f5gary: Ah, ok.

PR #389, opened 2017-10-16
CCCL incompatible with BIG-IP v11.6.1 policies

edarzins:
Incompatibility with policy rules fixed in f5-sdk 3.0.3, update
requirements to point to new CCCL version.


f5gary:
I think this commit should include an update to the attributions information, and probably a release note.

sjberman:
Yes, forgot that we wanted to include this in the release notes.

edarzins:
This was never broken in a release, only on "master" -- k8s v1.2 didn't have this bug.

sjberman:
Ah, I had thought this was an issue in the release, but you're right!

edarzins:
Where is the attribution info?

f5gary:
Try running "make docs" in your repo. I see the travis build does run that step, but I don't understand why it passes - I expect it to fail when the CCCL SHA changes since it is a dependency.

edarzins:
I ran 'make docs' and it created the correct attributions info for CCCL and the f5-sdk.

f5gary:
Alright, I guess there's nothing else left to do.

PR #391, opened 2017-10-17
Document partition management and configuration enforcement

### LINKED ISSUE #379: Document partition management and configuration enforcement
The CC will enforce orchestration configuration on the managed partition. Our meaning of "managed partition" includes the enforce of configuration; in other words, objects may be created, deleted, updated after administrator changes based on the authoritative configuration coming from the environment.

The consequences of this could mean disruptions of service and unexpected behavior.

Clearly document that administrator changes in the CC managed partition will result in BIG-IP configuration changes back to the state it expects.
NOTE: all CCs exhibit this behavior. Ensure consistency of language across each.

sjsharma2001:
Problem: The following is not clearly stated in the documentation. Administrator changes in the Controller managed partition will result in BIG-IP configuration changes back to the state the Controller expects.

Solution: Explicit highlight this in the documentation.

Fixes #379

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -58,6 +58,12 @@ For example, when run in `NodePort mode`_, the |kctlr| does the following:
 The BIG-IP system handles traffic for the Service at the specified virtual address and load balances to all nodes in the cluster.
 Within the cluster, the allocated NodePort load balances traffic to all pods.
 
+.. danger::
+ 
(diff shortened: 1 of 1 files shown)


sjsharma2001:
Hi @f5yacobucci, I have moved the text to the end of Overview section as you requested.

PR #393, opened 2017-10-20
Added non-0 route domain support

russokj:
Problem:  Need to make sure controllers handle Big-IP addresses that
          have optional route domain IDs specified.
Solution: Allowed virtual address field to contain route domains by
          updating schema to define Bigip IP formats that include
          route domains.

### DIFF
--- a/python/k8s-build-requirements.txt
+++ b/python/k8s-build-requirements.txt
@@ -1,6 +1,6 @@
--e git+https://github.com/f5devcentral/f5-cccl.git@9764f43b57271e8da31b78b8e14515825da8147a#egg=f5-cccl
+-e git+https://github.com/f5devcentral/f5-cccl.git@5edac6c7be8fb105cc3394d846949c329f5b829d#egg=f5-cccl
 pytest==3.0.2
 mock
-flake8
+flake8==3.4.1
 pytest-cov==2.5.1
 coveralls==1.1
--- a/python/k8s-runtime-requirements.txt
+++ b/python/k8s-runtime-requirements.txt
@@ -1,4 +1,4 @@
--e git+https://github.com/f5devcentral/f5-cccl.git@9764f43b57271e8da31b78b8e14515825da8147a#egg=f5-cccl
+-e git+https://github.com/f5devcentral/f5-cccl.git@5edac6c7be8fb105cc3394d846949c329f5b829d#egg=f5-cccl
 pyinotify==0.9.6
 ipaddress==1.0.17
 PyJWT==1.4.0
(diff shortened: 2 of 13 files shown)


schemas/bigip-virtual-server_v0.1.4.json
```diff
@@ -182,7 +182,7 @@
       "type": "object",
```
amudukutore: Question on this file: do we now require that the controller configs for the latest version require this schema file?  If so, it should be called out in the docs. The current docs list that v1.1.0 or higher require v0.1.3 of the schema.

schemas/test/util/schema-handler.js
```diff
@@ -17,6 +17,34 @@
...
 const fs = require('fs');
 const Validator = require('jsonschema').Validator;
+const ValidatorHelpers = require('jsonschema/lib/helpers');
+
+
+Validator.prototype.customFormats.bigipv4 = function(input) {
+   let parts = input.split("%");
+   let ip = parts[0];
+   if (parts.length === 2) {
+     letrd = parts[1];
```
edarzins: Should be "let rd", not "letrd".

python/k8s-build-requirements.txt
```diff
@@ -1,4 +1,4 @@
--e git+https://github.com/f5devcentral/f5-cccl.git@9764f43b57271e8da31b78b8e14515825da8147a#egg=f5-cccl
+-e git+https://github.com/f5devcentral/f5-cccl.git@e9b5144ef25e018714167582b12b802b3bc85f46#egg=f5-cccl
```
edarzins: f5-cccl SHA with latest changes is 5edac6c7be8fb105cc3394d846949c329f5b829d

python/k8s-runtime-requirements.txt
```diff
@@ -1,4 +1,4 @@
--e git+https://github.com/f5devcentral/f5-cccl.git@9764f43b57271e8da31b78b8e14515825da8147a#egg=f5-cccl
+-e git+https://github.com/f5devcentral/f5-cccl.git@e9b5144ef25e018714167582b12b802b3bc85f46#egg=f5-cccl
```
edarzins: f5-cccl SHA with latest changes is 5edac6c7be8fb105cc3394d846949c329f5b829d

schemas/bigip-virtual-server_v0.1.4.json
russokj: I believe from the comments on the v0.1.4 schema, the change for supporting multiple SSL profiles is backwards compatible.  The current change for supporting bigip style addresses is also backwards compatible. This was verified with the systests which currently use v0.1.3 of the schema.

python/k8s-build-requirements.txt
russokj: Updated with the latest SHA.

PR #407, opened 2017-10-27
Add slack channel info to SUPPORT.md

sjsharma2001:


amudukutore:
I've gone back and forth on this and am not sure if we should have this information on the top-level README page or in SUPPORT. My current preference is to have this in the top-level README under a Help subsection. Will let @f5yacobucci chime in with his thoughts.

sjsharma2001:
I am closing this, and will make this change in README.

PR #419, opened 2017-11-09
Add UDP support

edarzins:
Added UDP support via ConfigMaps and updated to CCCL version that
supports UDP Health Monitors.

### DIFF
--- a/pkg/appmanager/appManager_test.go
+++ b/pkg/appmanager/appManager_test.go
@@ -41,10 +41,12 @@ import (
 func init() {
 	workingDir, _ := os.Getwd()
 	schemaUrl = "file://" + workingDir + "/../../schemas/bigip-virtual-server_v0.1.4.json"
+	schemaUrl5 = "file://" + workingDir + "/../../schemas/bigip-virtual-server_v0.1.5.json"
 	DEFAULT_PARTITION = "velcro"
 }
 
 var schemaUrl string
+var schemaUrl5 string
 
 var configmapFoo string = string(`{
   "virtualServer": {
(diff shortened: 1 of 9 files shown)


schemas/bigip-virtual-server_v0.1.5.json
```diff
@@ -0,0 +1,212 @@
+{
+  "$schema": "http://json-schema/org/schema#",
+  "id": "f5schemadb://bigip-virtual-server_v0.1.4.json",
```
f5gary: Should be v0.1.5

pkg/appmanager/appManager_test.go
```diff
@@ -41,10 +41,12 @@ import (
 func init() {
 	workingDir, _ := os.Getwd()
 	schemaUrl = "file://" + workingDir + "/../../schemas/bigip-virtual-server_v0.1.4.json"
+	schemaUrl5 = "file://" + workingDir + "/../../schemas/bigip-virtual-server_v0.1.5.json"
```
f5gary: What are the differences with the v0.1.5 schema - is it not backwards compatible? I think you should be able to change schemaUrl to v0.1.5 and it would still work for other tests, so no need for a new variable.

edarzins: Adding 'udp' isn't a breaking change -- so should I just add the changes to v0.1.4 of the schema?

f5gary: No, it still needs to be a new version number as it is adding some optional new thing, but we should just use v0.1.5 in unit tests then. And now that I'm thinking of it, there should also be a change in docs/README.rst to include the new schema version and the changes it brings.

PR #421, opened 2017-11-13
Generalize VXLAN mgr; ARP support for Flannel

sjberman:
Problem: The current fdb record setup is named specifically for OpenShift SDN. Flannel support will reuse the same code, so naming should be more generalized. Flannel also requires ARP entries to be written to the BIG-IP system.

Solution: Generalize the naming of the openshift sdn package to be "vxlan". This will be used by both OpenShift SDN and Flannel. Add a CLI argument for the flannel vxlan tunnel name. ARP entries are also now configured and written out for Flannel.


sjberman:
This will need to wait to be merged until the CCCL changes are in, but I am opening this now just to get feedback ahead of time.

cmd/k8s-bigip-ctlr/main.go
```diff
@@ -102,14 +104,15 @@ var (
 	// package variables
 	isNodePort         bool
 	watchAllNamespaces bool
+	VXLAN_NAME         string
```
f5gary: What's the difference between VXLAN_NAME and openshiftSDNName above? Also, why is it not camelCase?

cmd/k8s-bigip-ctlr/main.go
```diff
@@ -321,8 +347,9 @@ func verifyArgs() error {
 func setupNodePolling(
 	appMgr *appmanager.Manager,
 	np pollers.Poller,
+	eventChan chan interface{},
```
f5gary: The channel parameter should have a direction unless it's bi-directional.

cmd/k8s-bigip-ctlr/main.go
sjberman: I can change it to be camelCase. This variable is used to represent either the flannel tunnel name or the openshift sdn name, for simplicity's sake later on in the code.

sjberman:
Addressed the comments, but I have one more CCCL change I'll have to pull in before this is merged.

PR #427, opened 2017-11-15
Create resource copy for debug logging

### LINKED ISSUE #365: Removing customProfiles from output causes errors in DEBUG
When writing the "resources" config blob out to the python driver, the controller removes customProfiles from this blob (after sending first) in order to avoid writing certs/keys out to the logs when in debug mode.

The problem with this is that when the controller is running in debug mode, the removal of profiles from the written config can cause the python driver to see an empty profiles blob on a second iteration, and attempt to delete the profiles off of the BIG-IP. The profiles won't be deleted, as they are being referenced by virtual servers, but error messages will be logged and can cause confusion for a user running in debug mode.

sjberman:
Problem: When debug logging, customProfiles were removed from the resources struct. This could cause future syncing issues because we removed a reference to the profiles, so our internal store was in an incorrect state.

Solution: Don't touch the resources blob when attempting to alter what is logged (customProfiles in this case). Instead, create a value copy of the resources into a new struct, leaving out the profiles. Log this new copied struct, which avoids us from altering our internal store.

Fixes #365

### DIFF
--- a/pkg/appmanager/outputConfig.go
+++ b/pkg/appmanager/outputConfig.go
@@ -87,6 +87,26 @@ func (appMgr *Manager) outputConfigLocked() {
 		}
 	}
 
+	for _, profile := range appMgr.customProfiles.profs {
+		initPartitionData(resources, profile.Partition)
+		resources[profile.Partition].CustomProfiles = append(resources[profile.Partition].CustomProfiles, profile)
+	}
+	for _, irule := range appMgr.irulesMap {
+		initPartitionData(resources, irule.Partition)
+		resources[irule.Partition].IRules = append(resources[irule.Partition].IRules, *irule)
+	}
+	for intDgKey, intDgMap := range appMgr.intDgMap {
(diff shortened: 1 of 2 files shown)


pkg/appmanager/outputConfig.go
```diff
@@ -125,18 +138,8 @@ func (appMgr *Manager) outputConfigLocked() {
...
-	}
-	for _, irule := range appMgr.irulesMap {
-		initPartitionData(resources, irule.Partition)
-		resources[irule.Partition].IRules = append(resources[irule.Partition].IRules, *irule)
-	}
-	for _, intDg := range appMgr.intDgMap {
-		initPartitionData(resources, intDg.Partition)
-		resources[intDg.Partition].InternalDataGroups = append(resources[intDg.Partition].InternalDataGroups, *intDg)
-	}
+	// Copy everything from resources except CustomProfiles to be used for debug logging
```
f5gary: This should be inside the if block at line 160 so the copy is only done when in debug mode.

PR #430, opened 2017-11-17
Ingress events for multiple namespaces are confusing

### LINKED ISSUE #417: Events published to Ingress resources in multiple namespaces result in confusing error messages
When the controller is handling Ingress resources in multiple namespaces, errors appear in the controller log after publishing events on the Ingress resources. This behavior has not yet been seen when all of the Ingress resources are within the same namespace.

E1107 01:50:59.704583       1 event.go:208] Unable to write event: 'can't create an event with namespace 'kube-system' in namespace 'testing'' (may retry after sleeping)
E1107 01:50:59.704585       1 event.go:208] Unable to write event: 'can't create an event with namespace 'testing' in namespace 'kube-system'' (may retry after sleeping)

f5gary:
Problem:
The controller was creating only one event broadcaster for all
namespaces, which is not correct. This causes errors when broadcasting
events once an event is generated for a second namespace.

Solution:
1. Added some new infrastructure to support one event broadcaster per
   namespace.
2. Moved all of the event notification code to eventNotifier.go.
3. Updated existing Ingress tests to validate events.
4. Added a new unit test to validate the new one event broadcaster per
   namespace functionality.
5. Updated some methods in mockAppManager for Ingress objects to update
   the fake kube client with Ingress objects to allow the correct code
   path for events to be followed (like it happens in production).
6. Some cleanup around the Params structure where it was allowing
   public access to members intended only for unit tests.

Fixes #417

affects-branches: master

### DIFF
--- a/pkg/appmanager/eventNotifier.go
+++ b/pkg/appmanager/eventNotifier.go
@@ -58,7 +58,8 @@ func NewEventNotifier(bfunc NewBroadcasterFunc) *EventNotifier {
 	}
 }
 
-func (en *EventNotifier) getNotifierForNamespace(
+// Create a notifier for a namespace, or return the existing one
+func (en *EventNotifier) createNotifierForNamespace(
 	namespace string,
 	coreIntf corev1.CoreV1Interface,
 ) *NamespaceEventNotifier {
@@ -68,10 +69,6 @@ func (en *EventNotifier) getNotifierForNamespace(
 
 	evNotifier, found := en.notifierMap[namespace]
 	if !found {
-		if nil == coreIntf {
-			// Don't create notifier for this namespace (unit testing)
-			return nil
-		}
 		source := v1.EventSource{Component: "k8s-bigip-ctlr"}
 		broadcaster := en.broadcasterFunc()
 		recorder := broadcaster.NewRecorder(scheme.Scheme, source)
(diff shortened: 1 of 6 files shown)


pkg/appmanager/eventNotifier.go
```diff
@@ -0,0 +1,133 @@
...
+	eventType,
+	reason,
+	message string,
+) {
+	nen.recorder.Event(obj, eventType, reason, message)
+}
+
+// This function expects either an Ingress resource or the name of a VS for
+// an Ingress.
+func (appMgr *Manager) recordIngressEvent(ing *v1beta1.Ingress,
```
dramich: Update formatting to match getNotifierForNamespace

pkg/appmanager/eventNotifier.go
```diff
@@ -0,0 +1,133 @@
...
+		// No broadcaster func provided (unit testing), use real one.
+		bfunc = record.NewBroadcaster
+	}
+	return &EventNotifier{
+		notifierMap:     make(map[string]*NamespaceEventNotifier),
+		broadcasterFunc: bfunc,
+	}
+}
+
+func (en *EventNotifier) getNotifierForNamespace(
```
dramich: I think this should be two functions, one to get and one to create. It does more than it says it does which could confuse a caller.

f5gary: I can rename it and create a second version for the unit test case that's simpler, but I don't want 2 functions for the normal usage path because then there's 2 locks/unlocks and a chance for something to happen in between. The get (or create and get) should occur with a single lock/unlock.

PR #440, opened 2017-11-29
Add version and build info to Docker container

edarzins:
Problem:
There are several places where the version info should be embedded, like the
controller binary, into the docs, and in docker image labels

Solution:
Add a version-tool that parses next-version.txt, as well as looking at git
and CI envrionment variables. Use this tool to add docker labels, add a
--version flag to the k8s-bigip-ctlr, dump version info to the controller
log, and append version info to the User-Agent header of iControl REST
requests.


michaeldayreads:
👍

PR #444, opened 2017-11-30
recv strings on health monitors

### LINKED ISSUE #275: Support recv strings on health checks
We currently allow users to specify the send string in a health check, but not the recv string.  We should support both.  CCCL supports [recv](https://github.com/f5devcentral/f5-cccl/blob/73f81e7baa6c515ba8b062e54c8d75de2624c424/f5_cccl/schemas/cccl-api-schema.yml#L418-L425), so I think this involves:

- Updating our [schema](https://github.com/F5Networks/k8s-bigip-ctlr/blob/cbb2a9585e152c2c2b0beadee55376c5957f0494/schemas/bigip-virtual-server_v0.1.4.json#L103) to have the recv type (optional).
- Updating the unit/system test for healthchecks to provide a recv string.
- Documenting the additional "recv" string.  It looks like we roughly document the health checks inside the ingress resource ("Each health monitor JSON object must have the following 4 fields...") but:
  - This looks incorrect, since only "protocol" is required at least according to the schema.
  - We don't document the health check types for the VirtualServer (and I think they are the same)
  - This example is nice, but it is in a different format than the rest of our documentation tables.

So I'd suggest making a section that correctly documents the format for a healthcheck and linking to it from both the VirtualServer and Ingress resources.

michaeldayreads:
Problem:
Controller currently does not support health monitoring based on an expected received string from a service.

Solution:
- Added 'recv' field support to config map, ingress, and route generated services.
- Added schema version 0.1.6 to include recv string

Tests:
- Added unit tests for controller code
- Added corresponding integration tests to nightly runs to confirm services are configured using passed recv string and a) show as healthy when response from server matches and b) shows as unhealthy when response from server does not match.
- Note: Schema tests added in separate commit to accomodate schema tests being behind one version.

fixes #275

---

Update schema tests for udp

Problem:
- schema tests were not updated for udp

Solution:
- added udp to expected error, rolled test version.

---

Schema tests for Health Monitor recv field
Problem:
- Schema tests did not validate recv field in health monitors.

Solution:
- Added schema tests for health monitor recv field and rolled schema version.


pkg/appmanager/healthMonitors_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/F5Networks/k8s-bigip-ctlr/pkg/test"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
```
sjberman: The build failed with this import. Is this actually used anywhere?

PR #459, opened 2017-12-13
Set InsecureEdgeTerminationPolicy affects all Route objects.

### LINKED ISSUE #341: Set insecureEdgeTerminationPolicy to Redirect in one edge route but affect all other edge route
Set insecureEdgeTerminationPolicy to Redirect in one edge route, it can work. But all other edge route even in different namespace also have the ability to redirect HTTP client connections to HTTPS.

If set insecureEdgeTerminationPolicy to Allow in one edge route, it only affect that route as expected.

f5gary:
Problem:
Since there is only one virtual for HTTP OpenShift Routes, any Route
that specifies that HTTP should be redirected to HTTPS affects all
Routes.

Solution:
1. The controller now maintains a new data group for HTTP redirects.
2. Updated the HTTP redirect IRule to perform a lookup to the new data
   group by host and path to determine if the HTTP request should be
   redirected.
3. The controller keeps the redirect data group updated for both
   OpenShift Route and Ingress resources to prevent the need to
   maintain 2 IRules for HTTP redirect and to automatically do the
   right thing when shared-ip for Ingress is implemented. This required
   some the existing data group handling code to be moved so it would
   be able to work across resource types (previously it was only used
   by Routes).
4. Removed some dead code from handleIngressTls left over from the
   first generation of HTTP redirect (policy based).
5. Discovered and fixed an issue with potential duplicate data when
   combining data group records from multiple namespaces.
6. Updated existing unit tests and wrote 2 new ones to test the new
   functionality for Routes and Ingresses.

Fixes #341

affects-branches: master

### DIFF
--- a/pkg/appmanager/resourceConfig.go
+++ b/pkg/appmanager/resourceConfig.go
@@ -819,6 +819,7 @@ func createRSConfigFromRoute(
 	routeConfig RouteConfig,
 	pStruct portStruct,
 	svcIndexer cache.Indexer,
+	svcFwdRulesMap ServiceFwdRuleMap,
 ) (ResourceConfig, error, Pool) {
 	var rsCfg ResourceConfig
 	rsCfg.MetaData.RouteProfs = make(map[routeKey]string)
@@ -831,7 +832,6 @@ func createRSConfigFromRoute(
 		policyName = "openshift_secure_routes"
 		rsName = routeConfig.HttpsVs
 	}
-	tls := route.Spec.TLS
 
 	var backendPort int32
 	var err error
@@ -899,9 +899,6 @@ func createRSConfigFromRoute(
 				}
 			}
 		}
-		if !found {
-			rsCfg.HandleRouteTls(tls, pStruct.protocol, policyName, rule)
-		}
 	} else { // This is a new VS for a Route
 		rsCfg.MetaData.ResourceType = "route"
 		rsCfg.Virtual.Name = rsName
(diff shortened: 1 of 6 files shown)


pkg/appmanager/resourceConfig.go
```diff
@@ -919,6 +994,18 @@ func createRSConfigFromRoute(
...
+		tls.InsecureEdgeTerminationPolicy == routeapi.InsecureEdgeTerminationPolicyRedirect {
+		// TLS config indicates to forward http to https.
+		path := "/"
+		if route.Spec.Path != "" {
+			path = route.Spec.Path
+		}
+		svcFwdRulesMap.AddEntry(route.ObjectMeta.Namespace, route.Spec.To.Name,
+			route.Spec.Host, path)
+	}
+
```
sjberman: Would this make more sense to put into HandleRouteTls?

pkg/appmanager/resourceConfig.go
```diff
@@ -1296,6 +1383,96 @@ func (idg *InternalDataGroup) RemoveRecord(name string) bool {
...
+		Partition: partition,
+		Name:      name,
+	}
+	for name, data := range flatMap {
+		newDg.AddOrUpdateRecord(name, data)
+	}
+
+	return &newDg
+}
+
```
sjberman: I think it may make more sense to take all of the data group related additions in this file and instead put them into routing.go, where the other data group processing lives. That includes everything in this file except the block in createRSConfigFromRoute where I left an earlier comment.

pkg/appmanager/resourceConfig.go
f5gary: HandleRouteTls isn't being called for updates (I'm not sure why) so if it was there we wouldn't handle changes to tls.InsecureEdgeTerminationPolicy properly.

pkg/appmanager/resourceConfig.go
f5gary: Will do.

PR #462, opened 2017-12-18
Shared Ingress support

### LINKED ISSUE #325: Single shared Virtual Server
Hi!

I've read the docs and tried a few different things, but can't seem to get it to work for our scenario. 
I'd like to be able to share **1 single Virtual Server** (1 IP) and have **k8s-bigip-ctlr** manage the pools and IRules or Policies to forward traffic to the correct pool. 

Using an _Ingress_ resource wouldn't work as it will create a new Virtual Server (which requires a new IP) per Ingress/Application. 
Name Based _Ingress_ _could_ work, but unfortunately, each app runs in its own namespace and must be able to provide their own _Ingress_ object (with the current k8s-bigip-ctlr implementation this would mean a new IP address per _Ingress_ object as VS can't be shared).

One thing that could work is having multiple _Ingress_ objects updating the same BIGP _Policy_ , but I doesn't seem like we can do this at the moment (BIGIP won't allow policies with conflicting controls associated with the same VS).

I found this in the docs:
_

> If virtualAddress or bindAddr are not provided in the Frontend configuration, then the controller will configure and manage pools, pool members, and healthchecks for the service without a virtual server on the BIG-IP. Instead you should already have a BIG-IP virtual server that handles client connections and has an irule or traffic policy to forward the request to the correct pool. The stable name of the pool will be the namespace of the Kubernetes service followed by an underscore followed by the name of the service ConfigMap.

_

This means that we could get this scenario working using ConfigMaps describing the pool, and then using another external system to create the IRule or traffic policy to forward the request to the correct pool. 

Is there any other way we could achieve this?
Are there any plans to support this scenario?

sjberman:
Add support for Ingresses with the same IP:Port specification to share virtual servers.

1. Pools for Ingresses have been renamed to exclude the name of the Ingress resource, that way multiple Ingresses using the same backend don't create new pools; the pools are shared.
2. Virtual server names are of the format "ingress_ip_port".
3. If Ingresses have different IP addresses then they will create separate virtual servers.

Closes #325

### DIFF
--- a/pkg/appmanager/resourceConfig.go
+++ b/pkg/appmanager/resourceConfig.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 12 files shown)


pkg/appmanager/resourceConfig.go
```diff
@@ -719,21 +749,18 @@ func copyConfigMap(virtualName, ns string, cfg *ResourceConfig, cfgMap *ConfigMa
 // Create a ResourceConfig based on an Ingress resource config
 func createRSConfigFromIngress(
 	ing *v1beta1.Ingress,
+	resources Resources,
```
f5gary: This should be a *Resources since it contains locks and shouldn't be copied.

sjberman: Good catch. I need to start keeping this in mind more now.

f5gary:
Also looks like Travis failed.

PR #467, opened 2017-12-20
WIP - Added pools for alternate backends in route definition

russokj:
Problem:  A/B deployments are possible in openshift by using the
          AlternateBackends field that specifies additional services
          and weights.  The k8s controller currently ignores the
          field

Solution: Phase 1 - Create additional pools for each service listed
          in the AlternateBackends field.

Note:  A follow on commit will be required to add an irule that
       allows the use of the additional services for the route.


f5gary:
I think Saylor and I have addressed your immediate concerns for this WIP PR in our slack conversation.

sjberman:
@russokj If you want to resolve the conflicts, I'll review after that.

PR #469, opened 2018-01-03
Process multiple services per route definition

russokj:
Problem:  Need to be able to handle the alternateBackends definition
in the route spec.  This is used to specify addition services to use
for A/B deployments

Solution: Updated the route handling to return an array of services
instead of just one.

### DIFF
--- a/pkg/appmanager/resourceConfig.go
+++ b/pkg/appmanager/resourceConfig.go
@@ -220,6 +220,29 @@ func formatIngressPoolName(namespace, ingName, svc string) string {
 	return fmt.Sprintf("ingress_%s_%s_%s", namespace, ingName, svc)
 }
 
+func getRouteCanonicalService(route *routeapi.Route) string {
+	return route.Spec.To.Name
+}
+
+// return the services associated with a route
+func getRouteServiceNames(route *routeapi.Route) []string {
+	numOfSvcs := 1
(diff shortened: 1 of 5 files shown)


pkg/appmanager/resourceConfig.go
```diff
@@ -220,6 +220,31 @@ func formatIngressPoolName(namespace, ingName, svc string) string {
...
+	svcs := make([]string, numOfSvcs)
+
+	svcIdx := 0
+	if route.Spec.AlternateBackends != nil {
+		for i, svc := range route.Spec.AlternateBackends {
+			svcIdx = i
+			svcs[svcIdx] = svc.Name
+		}
+	}
+	svcs[svcIdx] = getRouteCanonicalService(route)
```
sjberman: Won't this overwrite the last alternate backend in the list, since svcIdx isn't incremented once the loop has exited?

russokj: I believe it would.  I moved it from being the first to the last, so I forgot this important detail.  Good catch.

russokj: I just noticed this is an older version of the code.  I forgot to push the latest.  Will have a new version pushed in a minute or two.

sjberman:
Looks like just a gofmt error with that method. Looks good otherwise.

PR #474, opened 2018-01-08
Feature.ab deployment.alt backend pools

russokj:
WIP - Feature.ab_deployment branch is messed up, but the commit 'Added creation of pools for alternatedBackends' is at least reviewable (as well as the follow on commit based on review comments).

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -605,8 +605,10 @@ func (appMgr *Manager) enqueueIngress(obj interface{}) {
 }
 
 func (appMgr *Manager) enqueueRoute(obj interface{}) {
-	if ok, key := appMgr.checkValidRoute(obj); ok {
-		appMgr.vsQueue.Add(*key)
+	if ok, keys := appMgr.checkValidRoute(obj); ok {
+		for _, key := range keys {
+			appMgr.vsQueue.Add(*key)
+		}
 	}
 }
 
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -1110,6 +1110,14 @@ func (appMgr *Manager) syncRoutes(
 		if route.ObjectMeta.Namespace != sKey.Namespace {
 			continue
 		}
+
+		//FIXME(kenr): why do we process services that aren't associated
+		//             with a route?
+		svcName := route.Spec.To.Name
+		if existsRouteServiceName(route, sKey.ServiceName) {
+			svcName = sKey.ServiceName
+		}
+
 		if nil != route.Spec.TLS {
 			switch route.Spec.TLS.Termination {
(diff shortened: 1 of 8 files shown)


pkg/appmanager/appManager.go
```diff
@@ -1457,6 +1471,7 @@ func (appMgr *Manager) handleConfigForType(
 
 	deactivated := false
 	if _, ok := svcPortMap[pool.ServicePort]; !ok {
+			pool.ServiceName, svcKey.Namespace)
```
sjberman: This feels out of place.

pkg/appmanager/resourceConfig.go
```diff
@@ -980,6 +1016,7 @@ func createRSConfigFromRoute(
 		}
 		if !found {
 			rsCfg.Pools = append(rsCfg.Pools, pool)
+		} else {
```
sjberman: Empty else here.

pkg/appmanager/routing.go
```diff
@@ -394,7 +394,7 @@ func updateDataGroupForPassthroughRoute(
 	dgMap InternalDataGroupMap,
 ) {
 	hostName := route.Spec.Host
-	poolName := formatRoutePoolName(route)
+	poolName := formatRoutePoolName(route, route.Spec.To.Name) // TODO(kenr): Do we need to make this svcName? (same below)
```
sjberman: I think we would want this to be svcName. If a Route is going to specify more than one backend, then we want the passthrough/reencrypt rules to apply to any pools for that route. So the data group would need to know about each pool for the Route.

sjberman: It may be worth getting @f5gary's input on this as well.

pkg/appmanager/appManager.go
russokj: It is.  It was part of a print statement I forgot to completely remove.

pkg/appmanager/appManager_test.go
```diff
@@ -3179,11 +3179,12 @@ var _ = Describe("AppManager Tests", func() {
 					}
 					hostDg, found := mockMgr.appMgr.intDgMap[hostDgKey]
 					Expect(found).To(BeTrue())
+					// TODO(kenr): READD AFTER UNDERSTANDING WHY THESE FAILED
```
russokj: This comment is no longer needed.  The tests no longer fail.

sjberman:
Make sure you rebase on master, as the recent Shared Ingress work could have some impact on how this code will run (specifically around cleanup). There are also some conflicts.

russokj:
Closing out this review since the rebase wasn't properly done.  A new review is out that is against the feature branch which was updated with the latest from master (https://github.com/F5Networks/k8s-bigip-ctlr/pull/475).  That request should include the latest code changes.

PR #475, opened 2018-01-08
Added creation of pools for alternateBackends (openshift only)

russokj:
Problem: Each route may specify multipe services.  We need pools
created for each of these services.

Solution: During route sync, determine if the service belongs to
the route and process accordingly.

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -1110,6 +1110,17 @@ func (appMgr *Manager) syncRoutes(
 		if route.ObjectMeta.Namespace != sKey.Namespace {
 			continue
 		}
+
+		//FIXME(kenr): why do we process services that aren't associated
+		//             with a route?
+		svcName := getRouteCanonicalService(route)
+		if existsRouteServiceName(route, sKey.ServiceName) {
+			svcName = sKey.ServiceName
+		}
+
+		// Collect all service names for this Route.
+		svcs := getRouteServiceNames(route)
+
 		if nil != route.Spec.TLS {
 			switch route.Spec.TLS.Termination {
 			case routeapi.TLSTerminationPassthrough:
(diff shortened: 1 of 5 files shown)


sjberman:
Looks good to me. Just want to verify with @f5gary that we want to do the data group stuff for the alternate backends.

pkg/appmanager/appManager.go
```diff
@@ -1110,20 +1110,28 @@ func (appMgr *Manager) syncRoutes(
...
+		svcName := getRouteCanonicalService(route)
+		if existsRouteServiceName(route, sKey.ServiceName) {
+			svcName = sKey.ServiceName
+		}
+
 		if nil != route.Spec.TLS {
 			switch route.Spec.TLS.Termination {
 			case routeapi.TLSTerminationPassthrough:
-				updateDataGroupForPassthroughRoute(route, DEFAULT_PARTITION,
+				updateDataGroupForPassthroughRoute(route, svcName, DEFAULT_PARTITION,
```
f5gary: There is a problem with updating the data groups for one route with multiple services (which is what this commit is introducing). The problem is that the key in the data group is the host name defined in the route and the value is the pool name. The key is unique, so in this case the data group is updated for the same host multiple times - once for each backend, and the last update 'wins'.

pkg/appmanager/appManager.go
```diff
@@ -1110,20 +1110,28 @@ func (appMgr *Manager) syncRoutes(
...
 					sKey.Namespace, dgMap)
 			case routeapi.TLSTerminationReencrypt:
-				updateDataGroupForReencryptRoute(route, DEFAULT_PARTITION,
+				updateDataGroupForReencryptRoute(route, svcName, DEFAULT_PARTITION,
```
f5gary: Same comment for passthrough routes applies here as well for re-encrypt routes.

pkg/appmanager/appManager.go
russokj: For this commit, I will only update for the main (canonical) service in the route since this change is not doing A/B yet (just creating the extra pools).

pkg/appmanager/appManager.go
russokj: Same as above.

pkg/appmanager/appManager.go
```diff
@@ -1164,9 +1172,12 @@ func (appMgr *Manager) syncRoutes(
 				}
 			}
 
+			// Collect all service names for this Route.
+			svcs := getRouteServiceNames(route)
```
f5gary: This should move outside the for loop.

PR #488, opened 2018-01-17
Add A/B For TLS-based routes (plus several bug fixes)

russokj:
Adding A/B support for TLS-based deployments (passthrough is working; reencrypt is partially working).

### DIFF
--- a/pkg/appmanager/resourceConfig.go
+++ b/pkg/appmanager/resourceConfig.go
@@ -1051,16 +1051,6 @@ func createRSConfigFromRoute(
 		if !found {
 			rsCfg.Pools = append(rsCfg.Pools, pool)
 		}
-		// If rule already exists, update it; else add it
-		found = false
-		if len(rsCfg.Policies) > 0 {
-			for i, rl := range rsCfg.Policies[0].Rules {
-				if rl.Name == rule.Name || rl.FullURI == rule.FullURI {
-					found = true
-					rsCfg.Policies[0].Rules[i] = rule
-				}
-			}
-		}
 	} else { // This is a new VS for a Route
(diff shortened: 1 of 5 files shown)


pkg/appmanager/resourceConfig.go
```diff
@@ -1174,6 +1180,28 @@ func (rc *ResourceConfig) AddRuleToPolicy(
...
+	rule *Rule,
+) {
+	// We currently have at most 1 policy, 'forwarding'
+	policy := rc.FindPolicy("forwarding")
+	if nil != policy {
+		for i, r := range policy.Rules {
+			if r.Name == rule.Name && r.FullURI == rule.FullURI {
+				// Remove old rule
+				if len(policy.Rules) == 1 {
+					rc.DeletePolicy(*policy)
```
sjberman: Enhancement: You can take advantage of this new DeletePolicy method here: https://github.com/F5Networks/k8s-bigip-ctlr/blob/master/pkg/appmanager/appManager.go#L1721

pkg/appmanager/resourceConfig.go
```diff
@@ -1174,6 +1180,28 @@ func (rc *ResourceConfig) AddRuleToPolicy(
...
+				} else {
+					policy.Rules = append(policy.Rules[:i], policy.Rules[i+1:]...)
+					rc.SetPolicy(*policy)
```
sjberman: Check out the block here that deletes rules: https://github.com/F5Networks/k8s-bigip-ctlr/blob/master/pkg/appmanager/appManager.go#L1726. You'll have to do something similar with the offsets, and there is a method that specifically deletes rules. (The linked block deletes rules under a different condition, but the method of deleting should be applicable here)

PR #495, opened 2018-01-19
Add balance route annotation

### LINKED ISSUE #491: Add balance annotation to OpenShift Routes
#### Description
OpenShift Routes must support 'virtual-server.f5.com/balance' in parity with k8s Ingress resources.

CC supported LB algos should be scoped to these:
"fastest-app-response”,
“fastest-node”,
“least-connections-member”,
“least-connections-node”,
“least-sessions”,
“observed-member”,
“observed-node”,
“predictive-member”,
“predictive-node”,
“round-robin”,
"weighted-least-connections-member”,
"weighted-least-connections-node”

#### Kubernetes Version
OpenShift

#### Controller Version
1.3

#### BIG-IP Version
N/A

sjberman:
Problem: Routes would default to using round-robin as their load balancing method, without
the ability to configure a different method.

Solution: Routes can now be annotated with the 'virtual-server.f5.com/balance' annotation, similar
to Ingress, in order to use a different LB method than round robin.

Fixes #491

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -628,6 +628,8 @@ Supported Route Annotations
 +----------------------------------------+-------------+-----------+--------------------------------------------------------------------------------+-------------+
 | Annotation                             | Type        | Required  | Description                                                                    | Default     |
 +========================================+=============+===========+================================================================================+=============+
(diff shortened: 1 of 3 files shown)


sjberman:
Yes, let me add those.

PR #497, opened 2018-01-19
Clarify docs

### LINKED ISSUE #422: Documentation: Update CloudDocs allowed values to match schema
#### Description
This page http://clouddocs.f5.com/products/connectors/k8s-bigip-ctlr/v1.3/#configuration-parameters shows the allowed values for frontend.virtualserver mode to be round-robin only. Can we update it to have all the allowed values as defined in the current f5schemadb://bigip-virtual-server_v0.1.4.json.

jputrino:
Fixes #422 

This PR contains the agreed-upon fix for the schema-lb mode mismatch -- linking out to the schema in f5-cccl.

I have also made some general edits to improve organization and usability.

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -25,7 +25,7 @@ Features
 - Support for F5 `iApps`_.
 - Handles F5-specific VirtualServer objects created in Kubernetes.
 - Handles standard `Kubernetes Ingress`_ objects using F5-specific extensions.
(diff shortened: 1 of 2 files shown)


sjberman:
@jputrino FYI I just merged a change that adds the 'virtual-server.f5.com/balance' annotation as a supported Route annotation, so the "lb" footnote can be added to that as well.

PR #501, opened 2018-01-22
Unit tests for A/B deployment

sjberman:
Adding two new unit tests for A/B deployment. The first test verifies that alternate backends in Routes create the correct number of pools and rules. The second test verifies that the data groups are correctly populated with the weights and corresponding pool names (for each path).

Also fixed the PoolCount method to correctly count the number of pools (formerly it would count configs per svcKey, which resulted in duplicates in some cases).

### DIFF
--- a/pkg/appmanager/appManager_test.go
+++ b/pkg/appmanager/appManager_test.go
@@ -3043,7 +3043,7 @@ var _ = Describe("AppManager Tests", func() {
 						[]v1.ServicePort{{Name: "foo-80", Port: 80, NodePort: 37001}})
 					r = mockMgr.addService(fooSvc)
 					Expect(r).To(BeTrue(), "Service should be processed.")
-					Expect(resources.PoolCount()).To(Equal(2))
+					Expect(resources.PoolCount()).To(Equal(1))
 
 					rs, ok := resources.Get(
 						serviceKey{"foo", 80, "default"}, "https-ose-vserver")
@@ -3078,7 +3078,7 @@ var _ = Describe("AppManager Tests", func() {
 						[]v1.ServicePort{{Port: 80, NodePort: 37001}})
 					mockMgr.addService(barSvc)
 					Expect(r).To(BeTrue(), "Service should be processed.")
-					Expect(resources.PoolCount()).To(Equal(4))
+					Expect(resources.PoolCount()).To(Equal(2))
 
 					rs, ok = resources.Get(
 						serviceKey{"bar", 80, "default"}, "https-ose-vserver")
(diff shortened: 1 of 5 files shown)


pkg/appmanager/resourceConfig.go
```diff
@@ -257,7 +257,11 @@ func getRouteServices(route *routeapi.Route) []RouteService {
 	if route.Spec.AlternateBackends != nil {
 		for _, svc := range route.Spec.AlternateBackends {
 			svcs[svcIndex].name = svc.Name
-			svcs[svcIndex].weight = int(*(svc.Weight))
+			if svc.Weight != nil {
```
russokj: This is not possible in a real Openshift environment. Is this needed just for unit tests?

sjberman: If it's not possible in OpenShift, then I'll remove it and make sure the tests supply a weight.

pkg/appmanager/appManager_test.go
```diff
@@ -3443,6 +3443,176 @@ var _ = Describe("AppManager Tests", func() {
...
+					mockMgr.addService(bazSvc)
+
+					spec := routeapi.RouteSpec{
+						Host: "foo.com",
+						Path: "/foo",
+						To: routeapi.RouteTargetReference{
+							Kind: "Service",
+							Name: "fake1", // start with non-existent service
+						},
+						AlternateBackends: []routeapi.RouteTargetReference{
```
russokj: Just a FYI - If the version of openshift supports A/B, it will default to a weight of 100 for any service that does not specify a weight (including the 'To' service).  If openshift doesn't support A/B, the weight will be ignored (and you will never have a alternateBackends field.

pkg/appmanager/appManager_test.go
```diff
@@ -3443,6 +3443,176 @@ var _ = Describe("AppManager Tests", func() {
...
+						[]v1.ServicePort{{Port: 80, NodePort: 37001}})
+					mockMgr.addService(fooSvc)
+					barSvc := test.NewService(svc2Name, "1", namespace, "NodePort",
+						[]v1.ServicePort{{Port: 80, NodePort: 37002}})
+					mockMgr.addService(barSvc)
+					bazSvc := test.NewService(svc3Name, "1", namespace, "NodePort",
+						[]v1.ServicePort{{Port: 80, NodePort: 37003}})
+					mockMgr.addService(bazSvc)
+
+					wMax := int32(100)
```
russokj: Nit - not sure why this is called wMax.  100 isn't the max (I think 256 is).  In anycase, this seems to be the weight of the foo service, so that would be a better name (or w# like you have for the other weights).

pkg/appmanager/appManager_test.go
```diff
@@ -3443,6 +3443,176 @@ var _ = Describe("AppManager Tests", func() {
...
+					route2 := test.NewRoute("route2", "1", namespace, spec2, nil)
+					mockMgr.addRoute(route2)
+
+					// verify data groups
+					grpRef := nameRef{
+						Partition: DEFAULT_PARTITION,
+						Name:      abDeploymentDgName,
+					}
+					nsMap, found := mockMgr.appMgr.intDgMap[grpRef]
+					Expect(found).To(BeTrue(), "redirect group not found")
```
russokj: Is this error message correct?

pkg/appmanager/resourceConfig.go
russokj: It's not possible.

Issue #471, opened 2018-01-05
When updating routes, old service pools are not removed until after a refresh cycle


russokj:
#### Description
In Openshift, if a route changes which service it is using (for instance, upgrading to a new version), the pool for the old service it left configured on the BIG-IP until the next refresh cycle.  To test, configure two services in openshift: svc-v1 & svc-v2. Then configure a route to use svc-v1. When the route is changed to use svc-v2, the BIG-IP will show the pools for both svc-v1 and svc-v2 until either the next refresh cycle of the controller, or the user deletes the service svc-v1.

#### Kubernetes Version
Openshift v1.4.1

#### Controller Version
master as of 1/3/18

#### BIG-IP Version
BIG-IP 12.1.0 Build 1.0.1447 Hotfix HF1

#### Diagnostic Information
The following was the initial route definition before switch the service to svc-v2:
```
apiVersion: v1
kind: Route
metadata:
  name: foo-com
  namespace: default
spec:
  host: foo.com
  path: /foo
  to:
    kind: Service
    name: svc-v1
```

sjberman:
The root cause of this is that when a Route changes its Service, we see an update for that Route and grab the new Service off of the Route and move forward with processing. We don't have any information at that point about the previous Service that the Route pointed to, as the Route no longer contains that information, so we don't initially clean up the previous pool.

A potential fix for this is every time we process some update, we loop through all resources (ConfigMaps, Ingresses, Routes) and loop through all pools in our currently stored configs, and if any pools point to a Service that isn't in any of our resources, we delete those pools. This is expensive, though.

PR #518, opened 2018-01-25
Fix namespace bug when deleting profiles

### LINKED ISSUE #517: SSL profiles deleted if ctlr watching multiple namespaces
There is an issue where the controller will delete SSL profiles from a virtual server if it is watching multiple namespaces. This is do to a comparison problem where the ctlr looks for Ingresses in each namespace to verify if a virtual's profiles are referenced by these Ingresses. Looping through each namespace would end up giving a false negative since a profile wouldn't be referenced by an Ingress in a different namespace, therefore deleting it from the virtual.

#### Kubernetes Version
1.7

#### Controller Version
1.4

#### BIG-IP Version
12.1

sjberman:
Problem: When deleting unused profiles, if processing a namespace that did not contain
an Ingress that referenced a profile on a virtual, the controller would delete the profile (even
if the profile was referenced by an Ingress in a different namespace).

Solution: Add a namespace field to the ProfileRef struct, in order to tell which namespace a profile
reference comes from. This way we don't delete a profile uninentionally when processing a namespace that
the profile didn't come from.

Note: This is a temporary fix that can be properly reworked when the fix for bug #471  is in.

Fixes #517

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -1,6 +1,13 @@
 Release Notes for BIG-IP Controller for Kubernetes
 ==================================================
 
+v1.4.1
+------
+
+Bug Fixes
+`````````
+* :issues:`517` - Controller deletes SSL profiles off of Ingress virtual servers if watching multiple namespaces.
+
 v1.4.0
 ------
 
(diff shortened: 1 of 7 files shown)


sjberman:
I will also need to cherry-pick this into 1.4-stable.

PR #522, opened 2018-01-26
Old service pools are not immediately removed

### LINKED ISSUE #471: When updating routes, old service pools are not removed until after a refresh cycle

f5gary:
Problem:
When Route resources are updated to reference a different service, the
old pools are not cleaned up until the next full refresh cycle. This is
also applicable to Ingress resources.

Solution:
There was no way before this commit to detect that a Route or Ingress
has been updated to reference different services. Now the Resources
object has the ability to track dependencies for Ingress and Route
objects (currently only Service dependencies are tracked), which allows
changes in dependent services to be identified, and pools updated
accordingly.

This commit also contains a bug fix for deleteUnusedResources, fixing
an issue where a policy that contains only 1 rule is deleted regardless
if it is a rule that should be removed for that invocation. That
function was also not protected by the mutex as it should have been so
I added that and also removed the appInf parameter as it was not being
used.

Fixes #471

affects-branches: master

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -784,11 +784,12 @@ func (appMgr *Manager) processNextVirtualServer() bool {
 }
 
 type vsSyncStats struct {
-	vsFound   int
-	vsUpdated int
-	vsDeleted int
-	cpUpdated int
-	dgUpdated int
+	vsFound      int
+	vsUpdated    int
+	vsDeleted    int
+	cpUpdated    int
+	dgUpdated    int
+	poolsUpdated int
 }
 
 func (appMgr *Manager) syncVirtualServer(sKey serviceQueueKey) error {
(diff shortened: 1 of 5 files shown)


amudukutore:
@f5gary - can you please cherry-pick this into the 1.4-stable branch as well? Thanks.

Issue #441, opened 2017-11-29
Schema DB path is hardcoded


johscheuer:
#### Description
Currently the schema DB path is hardcoded in https://github.com/F5Networks/k8s-bigip-ctlr/blob/master/pkg/appmanager/resourceConfig.go#L56 this makes it hard to execute the F5 BigIP CTLR directly without a container. There are two options currently:
1.) modify the codebase
2.) link the schemadb to the needed path

I would recommend to make this path available as an parameter (e.g. like the pythonBaseDir).

PR #525, opened 2018-01-29
Fixes 441

johscheuer:
This resolves: https://github.com/F5Networks/k8s-bigip-ctlr/issues/441 and makes it possible to set  the `schemaLocal ` over the CLI.

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -113,6 +113,8 @@ type Manager struct {
 	useSecrets bool
 	// Channel for emitting events
 	eventChan chan interface{}
+	// Where the schemas reside locally
+	schemaLocal string
 }
 
 // Struct to allow NewManager to receive all or only specific parameters.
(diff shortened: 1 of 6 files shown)


pkg/appmanager/appManager.go
```diff
@@ -906,7 +910,7 @@ func (appMgr *Manager) syncConfigMaps(
 		if cm.ObjectMeta.Namespace != sKey.Namespace {
 			continue
 		}
-		rsCfg, err := parseConfigMap(cm)
+		rsCfg, err := appMgr.parseConfigMap(cm)
```
sjberman: This may just be personal preference, but I think I would have rather passed the schemaLocal string to this function, rather than turning it into a class method.

Do you have a reason you went this route?

johscheuer: I will change the according part of the code.

This was my first idea and since the function/method is only used by the AppMgr I thought it would be the best to make it a struct/class method

johscheuer:
Updated the code

Issue #510, opened 2018-01-24
Change behavior of controller on startup when encountering errors


russokj:
#### Description
The controller currently exits if the cccl driver (bigipconfigdriver.py) encounters any exceptions during startup.  It should ignore transient errors just like it would during the main processing loop.

#### Kubernetes Version
1.4

#### Controller Version
1.4
#### BIG-IP Version
12.1

#### Diagnostics
See #438 for further details.

PR #548, opened 2018-02-05
Add retry backoff for recoverable start up errors.

ryan-talley:
Problem:
 The bigipconfigdriver could encounter errors on start up that are
 recoverable but are currently fatal if encountered.

Solution:
 If the bigipconfigdriver encounters BIG-IP connection erros or missing
 config file errors (since these are written out by the controllers) it
 will now log these errors and enter a retry backoff loop until
 recovered.

Fixes: #510

### DIFF
--- a/python/bigipconfigdriver.py
+++ b/python/bigipconfigdriver.py
@@ -595,16 +595,20 @@ class ConfigWatcher(pyinotify.ProcessEvent):
 
 
 def _parse_config(config_file):
-    if os.path.exists(config_file):
-        with open(config_file, 'r') as config:
-            fcntl.lockf(config.fileno(), fcntl.LOCK_SH, 0, 0, 0)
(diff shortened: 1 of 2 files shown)


python/bigipconfigdriver.py
```diff
@@ -595,16 +595,20 @@ def process_default(self, event):
...
-            config_json = json.loads(data)
-            log.debug('loaded configuration file successfully')
-            return config_json
-    else:
-        return None
+    def _file_exist_cb():
+        if os.path.exists(config_file):
+            return (True, None)
+        else:
+            return (False, 'Waiting for config file {}'.format(config_file))
```
russokj: Who is using the second return value?

russokj: Never mind about my previous comment

PR #553, opened 2018-02-08
Merge pull request #550 from sjberman/annotation_addr_port

sjberman:
Fix an issue with vserver getting port 0


sjberman:
Build failed due to next-version not being updated, but that will be updated in a later commit.

PR #559, opened 2018-02-12
Clarify persistence of iApp limitations

recursivelycurious:
Problem:
Some iApps have limitations that prevent deployment by the controller - e.g. the Application Services Integration iApp is not compatible with BIG-IP > 12.1.1. Nightly regressions skip BIG-IP / Controller combinations that exceed these limitations and the documentation is not explicit.

Solution:
Add a base limitation to the release notes.

See https://github.com/F5Networks/f5-application-services-integration-iApp/issues/16

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -61,6 +61,7 @@ Limitations
   - `Download and install the latest iApps templates`_.
   - `Set the service to use the newer iApp template`_.
 
+* Check BIG-IP version compatibility on Application Services (iApps) before deploying. See Application Services Integration iApp `[#16] <https://github.com/F5Networks/f5-application-services-integration-iApp/issues/16>`_ for more information.
 * Cannot delete ARP entries on BIG-IP v11.6.1 when running the Controller in Kubernetes with Flannel VXLAN enabled.
 * The controller will exit at startup if it cannot establish a connection with the BIG-IP.
 


docs/RELEASE-NOTES.rst
```diff
@@ -181,6 +181,7 @@ Limitations
 * Parameters other than IPAddress and Port (e.g. Connection Limit) specified in the iApp Pool Member Table apply to all members of the pool.
 * Cannot configure virtual servers with IPv6 addresses in the configmap.
 * The K8S BIG-IP Controller cannot watch more than one namespace.
+* Application Services (iApp) limitations apply (e.g. the K8S BIG-IP Controller is compatible with BIG-IP 13.0 but the Application Servies Integration iApp is not. See Application Services Integration iApp `[#16] <https://github.com/F5Networks/f5-application-services-integration-iApp/issues/16>`_)
```
amudukutore: Couple of comments/questions on this:
1. This should be listed in the Limitations section for the latest released version of the controller (I'd be okay with any of 1.4.0, 1.4.1, or 1.4.2) unless you've only tested against 1.0.0 (where you've added this).
2. Not sure of this but the limitation listed here seems different than the one already called out in the Limitations section of 1.4.0. I was wondering if they were related. If so, perhaps these notes can be combined in a way that makes sense.

jputrino: In general, we avoid using abbreviations like "e.g.". Use "for example" instead.

Suggest adding the following bullet point to the Limitations section for v1.4.0:
```
- Applications Services (iApp) availability may be limited. Check BIG-IP version compatibility before deploying. See Application Services Integration iApp `[#16] <https://github.com/F5Networks/f5-application-services-integration-iApp/issues/16>`_ for more information.
```

recursivelycurious: @amudukutore: 
- the limitations are distinct. The [Application Services iApp](https://github.com/F5Networks/f5-application-services-integration-iApp) is not distributed through the download/release candidate package referenced in the limitation note regarding Route Domains for http and tcp iApps.
- The placement of the limitation in 1.0.0 was because the limitation has been present since that time, and the limitation was exposed as the test matrix expanded. I will move / update per your and @jputrino 's feedback.

Thanks!

recursivelycurious:
@jputrino - Could you give a quick review on the revision to please write-good?

PR #569, opened 2018-02-15
Replaced python with f5-ctlr-agent.

ryan-talley:
Problem:
 Current python dir was copied across the k8s-bigip-ctlr and the
 cf-bigip-ctlr projects which caused its code to unintentionally diverge
 between the two.

Solution:
 Switch k8s-bigip-ctlr to use the f5-ctlr-agent which is a single pip
 installable github repo that holds the code that was in the python dir.
 This will prevent the code from diverging in the future.


build-tools/build-release-images.sh
```diff
@@ -20,14 +20,12 @@ VERSION_INFO=$(${CURDIR}/version-tool version)
...
-mkdir -p $WKDIR/python
-cp python/*.py $WKDIR/python/
-cp python/k8s-runtime-requirements.txt $WKDIR/
+cp requirements.txt $WKDIR/
 cp schemas/bigip-virtual-server_v*.json $WKDIR/
 cp LICENSE $WKDIR/
 cp $CURDIR/help.md $WKDIR/help.md
 echo "{\"version\": \"${VERSION_INFO}\", \"build\": \"${BUILD_INFO}\"}" \
-  > $WKDIR/python/VERSION_BUILD.json 
+  > $WKDIR/VERSION_BUILD.json
```
sjberman: I think this is only used in the bigipconfigdriver, so I doubt you need to keep it here anymore.

ryan-talley: ya let me check how the driver actually uses this. Then I'll determine if it should be pulled out or not good catch though.

ryan-talley: This needs to stay but I am going to have to update the f5-ctlr-agent and then the git SHAs

requirements.txt
```diff
@@ -0,0 +1 @@
+-e git+https://github.com/f5devcentral/f5-ctlr-agent.git@1a9cca04a5bdc45af82c31ef9999d0b65be9f2cc#egg=f5-ctlr-agent
```
russokj: Why f5devcentral instead of f5networks?

cmd/k8s-bigip-ctlr/main.go
```diff
@@ -553,7 +550,7 @@ func main() {
 		BigIPPartitions: *bigIPPartitions,
 	}
 
-	subPidCh, err := startPythonDriver(configWriter, gs, bs, *pythonBaseDir)
+	subPidCh, err := startPythonDriver(configWriter, gs, bs)
```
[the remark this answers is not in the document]
ryan-talley: That is a good point. Another solution would be to make it an optional. If it is set then run 'python %s', pythonBaseDIr if not then run the pip installed version and log that pythonBaseDir is deprecated and should use a pip installed version instead.

requirements.txt
ryan-talley: This was from @f5yacobucci since this is more of an internal project similar to why CCCL is also f5devcentral.

cmd/k8s-bigip-ctlr/main.go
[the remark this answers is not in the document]
russokj: Sounds good.  Might even be useful for internal testing by developers.

ryan-talley:
FYI these are my final set of changes until https://github.com/f5devcentral/f5-ctlr-agent/pull/7 is merged then I will need to update the pointer to the f5-ctlr-agent then this will be ready to merge.

ryan-talley:
I have updated the f5-ctlr-agent pointer these are ready to merge on final approval.

PR #580, opened 2018-02-21
Cleanup for leftover objects (Ingress/Routes)

### LINKED ISSUE #567: [BUG]Openshift - cc does not delete objects it created
#### Description
<For bug: Describe the bug in detail, observed versus expected behavior, steps to reproduce the issue, and a brief description of your deployment scenario>
<For enhancement: Describe the enhancement request in detail>
When we use CC to handle Openshift routes, it will create all the needed objects on BIG-IP. However if we delete these routes, some objects will be deleted while some others will remain. It may be good to make sure that CC handle the whole lifecycle on any objects it creates on the BIG-IP

Example of objects that cc will not remove when a configmap / route / ab deployment is deleted: 

- if there is no more route, we would expect that the default http/https vs would be deleted
- nodes will remain  
- the iRule to spread the traffic between two application in an ab deployment remains. Its related datagroup remains also
- http_redirect_irule remains
- openshift_passthrough_irule remains
- https_redirect_dg datagroup
- ssl_passthrough_servername_dg datagroup
- ssl_reencrypt_servername_dg datagroup
- ssl_reencrypt_serverssl_dg datagroup
- Any monitors created through routes setup 

#### Kubernetes Version
<Version of Kubernetes being used>
oc v3.7.0+7ed6862
kubernetes v1.7.6+a08f5eeb62

#### Controller Version
<Version of k8s-bigip-ctlr being used>
1.4

#### BIG-IP Version
<Version of BIG-IP being used>
12.1.1
#### Diagnostic Information

Here is an example of my /config/partitions/kubernetes/bigip.conf after deployed a confimap, routes, ab deployment and then deleted those items. It would be expected that it's empty. 

[bigip.conf.gz](https://github.com/F5Networks/k8s-bigip-ctlr/files/1724526/bigip.conf.gz)

sjberman:
Problem: iRules and datagroups were being created early and left behind for both Ingresses and Routes.
The ctlr also had issues leaving behind health monitors and profiles for Routes.

Solution: Ensured that iRules and datagroups are only created when necessary, and are cleaned up when done being used. Also added some logic to clean up health monitors and profiles properly for Routes.

Fixes #567

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -191,28 +191,6 @@ func NewManager(params *Params) *Manager {
 	return &manager
 }
 
-func (appMgr *Manager) addIRule(name, partition, rule string) {
-	appMgr.irulesMutex.Lock()
-	defer appMgr.irulesMutex.Unlock()
-
-	key := nameRef{
-		Name:      name,
-		Partition: partition,
-	}
-	appMgr.irulesMap[key] = NewIRule(name, partition, rule)
-}
-
-func (appMgr *Manager) addInternalDataGroup(name, partition string) {
-	appMgr.intDgMutex.Lock()
-	defer appMgr.intDgMutex.Unlock()
-
-	key := nameRef{
-		Name:      name,
-		Partition: partition,
-	}
-	appMgr.intDgMap[key] = make(DataGroupNamespaceMap)
-}
-
 func (appMgr *Manager) watchingAllNamespacesLocked() bool {
 	if 0 == len(appMgr.appInformers) {
 		// Not watching any namespaces.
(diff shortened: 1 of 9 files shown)


pkg/appmanager/appManager.go
```diff
@@ -857,9 +820,10 @@ func (appMgr *Manager) syncVirtualServer(sKey serviceQueueKey) error {
 			return err
 		}
 	}
-
 	// Update internal data groups if changed
-	appMgr.updateRouteDataGroups(&stats, dgMap, sKey.Namespace)
+	appMgr.syncDataGroups(&stats, dgMap, sKey.Namespace)
+	// Delete IRules if necessary
```
russokj: Is this comment still accurate?

pkg/appmanager/resourceConfig.go
```diff
@@ -1543,6 +1695,21 @@ func (rc *ResourceConfig) RemovePool(
 			cfgChanged = true
 		}
 	}
+	// Delete health monitor for the pool
+	monitorName := fullPoolName + "_0_http"
```
sjberman: Turn this into a common function for naming monitors.

PR #595, opened 2018-03-01
Snat pool config

ryan-talley:


docs/README.rst
```diff
@@ -516,48 +535,51 @@ You can use the |kctlr| to `Expose Services to External Traffic using Ingresses`
...
+|    |                                          |             | [#hm1]_   |                                                                                     |             |                                         |
++----+------------------------------------------+-------------+-----------+-------------------------------------------------------------------------------------+-------------+-----------------------------------------+
+|    | recv                                     | string      | Optional  | String or RegEx pattern to match in first 5,120 bytes of backend response.          | N/A         |                                         |
++----+------------------------------------------+-------------+-----------+-------------------------------------------------------------------------------------+-------------+-----------------------------------------+
+| virtual-server.f5.com/source-addr-translation | string      | Optional  | Source address translation type to apply to the virtual server.                     | automap     | automap, none, snat                     |
```
russokj: how is the snat pool specified (same comment for routes too)?

ryan-talley: The full annotation would be: `"virtual-server.f5.com/source-addr-translation": "{"type":"snat", "pool":"snat-pool-name"}"`

ryan-talley:
Adding @jputrino could you please review my doc changes? Thanks.

pkg/appmanager/resourceConfig.go
```diff
@@ -813,6 +814,35 @@ func setProfilesForMode(mode string, cfg *ResourceConfig) {
 	}
 }
 
+func verifySourceAddrTranslation(vsName string, sat SourceAddrTranslation) error {
+	if sat.Type == "none" && sat.Pool != "" || sat.Type == DEFAULT_SOURCE_ADDR_TRANSLATION && sat.Pool != "" {
```
russokj: Use parentheses to make this clearer even though it's logically correct as is (to separate the and/or clauses).  Also, why is there a definition for "automap" but not for "none"? I'd create constants for all three possibilities (none, automap, snat).  There are idiomatic ways to create enum-like constants in GO.  You can then set the default to the automap contstant.

russokj: Wouldn't the logic
```
if sat.Type != "snat" && sat.Pool != "" {
```
be simpler?

docs/README.rst
```diff
@@ -90,7 +90,7 @@ General
 |                       |         |          |                                  | to poll the cluster for its             |                |
 |                       |         |          |                                  | node members.                           |                |
 +-----------------------+---------+----------+----------------------------------+-----------------------------------------+----------------+
-| python-basedir        | string  | Optional | /app/python                      | Path to python utilities                |                |
+| python-basedir        | string  | Optional | n/a                              | Path to python utilities                |                |
```
jputrino: Is this something that users don't need to configure? If so, maybe we should consider taking this row out.

docs/README.rst
```diff
@@ -368,6 +386,8 @@ sslProfile [#ssl]_   JSON object       Optional                   BIG-IP SSL pro
 
    You can also `assign IP addresses to BIG-IP virtual servers using IPAM`_.
 
+   See `Source Address Translation Overview`_ for more information on configuring virtual server source address translation.
```
jputrino: Change to "See the `Source Address Translation Overview`_ on AskF5 ..."

schemas/bigip-virtual-server_v0.1.8.json
```diff
@@ -0,0 +1,203 @@
...
+        "balance": { "type": "string", "minLength": 1},
+        "partition": { "type": "string", "minLength": 1 },
+        "mode": { "type": "string", "enum": [ "http", "tcp", "udp" ] },
+        "sslProfile": { "$ref": "#/definitions/sslProfileType" },
+        "virtualAddress": { "$ref": "#/definitions/virtualAddressType" },
+        "sourceAddressTranslation": {
+          "type": "object",
+          "properties": {
+            "type": { "type": "string", "enum": [ "automap", "none", "snat" ] },
+            "pool": { "type": "string", "minLength": 1}
```
russokj: Just checking to make sure minLength=1 is correct (for automap/none settings).

pkg/appmanager/resourceConfig.go
ryan-talley: Yes it would fixing.

docs/README.rst
ryan-talley: Changed.

schemas/bigip-virtual-server_v0.1.8.json
ryan-talley: This field should be left out when configuring automap/none that's why only types is required. So if you do add pool to those types that is technically not correct but we will ignore it and create the vs anyway and log about it.

docs/README.rst
ryan-talley: Ya this is something that they do not need to configure, but can optionally configure if they want to use their own python bigipconfigdriver.py instead of the one that is pip installed with the controller. So I'll defer to you whether this row needs to be deleted or not.

jputrino: Let's leave it in, then, just make sure it's clear what the intended use is (i.e., what you just explained to me).

docs/README.rst
```diff
@@ -642,39 +665,42 @@ Supported Route Configurations
...
+|                                               |             |           | terminated routes.                                                                |             |                                         |
++-----------------------------------------------+-------------+-----------+-----------------------------------------------------------------------------------+-------------+-----------------------------------------+
+| virtual-server.f5.com/source-addr-translation | string      | Optional  | Source address translation type to apply to the virtual server.                   | automap     | automap, none, snat                     |
+|                                               |             |           | JSON object encoded string see virtualServer table entry for formatting examples. |             |                                         |
++-----------------------------------------------+-------------+-----------+-----------------------------------------------------------------------------------+-------------+-----------------------------------------+
 
 Please see the example configuration files for more details.
```
jputrino: Please provide examples of the correct usage in the example config file(s).

ryan-talley: Added examples.

ryan-talley:
I'm holding off on merging until I get some more negative unit tests in as per @f5yacobucci suggestions.

pkg/appmanager/resourceConfig_test.go
```diff
@@ -659,6 +662,41 @@ var _ = Describe("Resource Config Tests", func() {
 				Expect(cfg.Virtual.VirtualAddress.BindAddr).To(Equal("5.6.7.8"))
 			})
 
+			It("handles invalid source address translation annotations for ingress", func() {
```
ryan-talley: @russokj here is the garbage input unit test I added as per Booch's comment yesterday feel free to review at your leisure.

pkg/appmanager/resourceConfig_test.go
```diff
@@ -717,12 +760,56 @@ var _ = Describe("Resource Config Tests", func() {
...
+				Expect(cfg.Virtual.SourceAddrTranslation.Type).To(Equal("none"))
+				Expect(cfg.Virtual.SourceAddrTranslation.Pool).To(Equal(""))
 				Expect(cfg.Pools[0].Name).To(Equal("openshift_default_bar"))
 				Expect(cfg.Pools[0].ServiceName).To(Equal("bar"))
 				Expect(cfg.Pools[0].ServicePort).To(Equal(int32(80)))
 				Expect(cfg.Policies[0].Name).To(Equal("openshift_insecure_routes"))
 				Expect(cfg.Policies[0].Rules[0].Name).To(Equal("openshift_route_default_route2"))
 			})
+
+			It("handles invalid source address translation annotations for routes", func() {
```
ryan-talley: Ditto 665.

Issue #433, opened 2017-11-18
Enhancement: Add Serversslprofile to Ingress integration


sennerholm:
#### Description
We have the requirement to encrypt both the incoming traffic to the Bigip vip and between the Bigip and the backends/pods in Kubernetes/Openshift (reencrypt in the bigip).
![Example_Of_setup](https://photos.app.goo.gl/9zVe0Bkx86EXU1Iv2)

We want to archieve that with an Ingress rules because we also have external firewalls which restrict the traffic depending on the ip address on the VIP (so we can’t use the router functionality). What we want is to have a way to fill the Server SSL Profile (Server) box.
![SSL_Profile_IN_Bigip](https://photos.app.goo.gl/tNMuf5mfWQ2hv0w12)
 
We can’t find that functionality in the current implementation of the f5 ctrl and propose something like this to be implemented:
 
Add a annotation for the serverssl in the ingress in the virtual-server.f5.com “namespace”, for example something like this:

Start of example:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: ingresssslrencrypt
  annotations:
    # Provide an IP address from the external VLAN on your BIG-IP device
    virtual-server.f5.com/ip: "10.11.12.13"
    # Specify the BIG-IP partition containing the virtual server
    virtual-server.f5.com/partition: "k8s"
    # NEW Proposed functionality trying to set server profile
    virtual-server.f5.com/serverssl: "/Common/serverssl-k8s"
    # Allow/deny TLS connections
    ingress.kubernetes.io/ssl-redirect: "false"
    # Allow/deny HTTP connections
    ingress.kubernetes.io/allow-http: "true"
spec:
  backend:
    # The name of the Kubernetes Service you want to expose to external traffic
    serviceName: securemonolith
    servicePort: 443
  tls:
    - secretName: monolithcert
```
End of example.
 
Sincerely
Mikael

#### Openshift Version
Openshift Origin 3.6, kubernetes 1.6

#### Controller Version
1.3

sjsharma2001:
Hi @sennerholm, thanks for filing the enhancement request. FYI, the two links in the description are not working.

f5gary:
Hi @sennerholm. We currently have this enhancement on our roadmap. We do support re-encrypt via OpenShift Route resources, but not with Kubernetes Ingress resources or our ConfigMap solution.

PR #606, opened 2018-03-07
Fix: #433 Add Serversslprofile to Ingress integration

sennerholm:
Problem:
We need to reencrypt from the BigIP to the PODs in Openshift

Solution:
Adding support for a new annotation in Ingress (from the #433 issue)

    virtual-server.f5.com/serverssl: "/Common/serverssl-k8s"

Detailed description of your solution:

Adding handling for serverssl in resourceconfig.go/handleIngressTls to add the profile
and profiles.go/deleteUnusedProfiles to don't accidently delete it. I also added a debug output so it's possible to logg unused profiles

Testing :
We have tested it on 1.4.3 branch in our environment based on Openshift 3.6 and 3.7.

### DIFF
--- a/pkg/appmanager/profiles.go
+++ b/pkg/appmanager/profiles.go
@@ -493,6 +493,16 @@ func (appMgr *Manager) deleteUnusedProfiles(
 							&referenced,
 						)
 					}
+					if serverProfile, ok :=
+						ing.ObjectMeta.Annotations[f5ServerSslProfileAnnotation]; ok == true {
+						appMgr.checkProfile(
+							prof,
+							&toRemove,
+							ing.ObjectMeta.Namespace,
+							serverProfile,
+							&referenced,
+						)
+					}
 					if referenced {
 						break
 					}
@@ -521,6 +531,8 @@ func (appMgr *Manager) deleteUnusedProfiles(
 				}
 			}
 			if !referenced {
+				log.Debugf("deleteUnusedProfiles Removing profile: %v.",
+					prof)
 				toRemove = append(toRemove, prof)
 			}
 		}
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -585,6 +585,8 @@ Supported Ingress Annotations
 | ingress.kubernetes.io/ssl-redirect            | boolean     | Optional  | Tells the Controller to redirect HTTP traffic to the HTTPS port for HTTPS Ingress   | true        | "true", "false"                         |
 |                                               |             |           | resources (see TLS Ingress resources, below).                                       |             |                                         |
 +-----------------------------------------------+-------------+-----------+-------------------------------------------------------------------------------------+-------------+-----------------------------------------+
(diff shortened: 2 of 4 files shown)


sennerholm:
Should I add an update to the documentation in the pull request?

sjberman:
Yes, please update the documentation with the new annotation. If you could also add a small unit test case (can add to an existing test if it makes sense, and you can use the Route reencrypt test as an example) just to verify that the profile is properly configured when this annotation is specified. Thanks!

pkg/appmanager/profiles.go
```diff
@@ -474,6 +474,22 @@ func (appMgr *Manager) deleteUnusedProfiles(
...
+						// If another slash remains, then we know a partition is in the name
+						if strings.ContainsAny(secretName, "/") {
+							profName = fmt.Sprintf("%s/%s", prof.Partition, prof.Name)
+						} else {
+							profName = prof.Name
+						}
+						if profName == secretName {
+							referenced = true
+						}
+					}
```
sjberman: A commit just went in that took this common code block and moved it into a function (also pulled in the "secrets" check from above). If you want to rebase on the latest commit, you could likely remove this chunk and use that function, as it fits with your annotation check.

sennerholm:
Updated according to @sjberman change

amudukutore:
@sennerholm - Thanks for contributing this enhancement. 

We have a feature on our roadmap for supporting server-side SSL profiles (to be delivered as part of our next cycle, i.e., in ~3-4 months). As part of this feature, we were planning to add support for:
i. ability to attach existing SSL profiles on the BIG-IP to an Ingress resource as a server-side profile
ii. ability to attach a Kubernetes secret to an Ingress resource as a server-side profile
iii. ability to attach existing SSL profiles on the BIG-IP to a ConfigMap resource as a server-side profile
iv. ability to attach a Kubernetes secret to a ConfigMap resource as a server-side profile

A couple of follow-on comments/questions for you:
1. Your current changes cover requirement (i) above. Are requirements (ii), (iii), (iv) on your radar? If so, would you be willing to help out with those?
2. Your changes, once merged, will need to run through our existing regression tests (run in our internal test-lab infrastructure). If we see new regressions due to this change, we may get back to you to make additional fixes.
3. Your enhancement will go out as part of our upcoming feature release (v1.5.0) which should be out in about a month.

sennerholm:
Hi @amudukutore 

1) In my customers environment we have a requirement for (i), (ii) is nice to have, and I may have the time to look into it in the future, but first I need to get some (i) in and get the rest of the environment up and running. (iii) and (iv) we doesn't have on the radar at the moment, but maybe when you have made an integration against an IPAM. I can explain more on slack.

2) Let me know if you need something more or find some regression. I have implemented the changes @sjberman requested before. I see that the formatting in the README is updated, I doesn't have the time to fix it today (it's late here in europe), but I will try to update the pull request tomorrow so it's possible to merge. 

3) That would be great, I will probably do a backport to a local 1.4 as soon as you have merged it wish I can use until v1.5.0 is released.

mzahorik:
+1 to releasing sooner rather than later, even if that means a 1.4.x point release.

Issue #619, opened 2018-03-15
Move SNAT sample configs to separate configs.


ryan-talley:
#### Description
Currently the some of the configmap/ingress/route config examples include SNAT pool configurations which may be confusing/break customer configs if they were trying to use those configs blindly. These should be moved into configmap /ingress/route sample configs that are separate to avoid conflicts.

#### Controller Version
v1.5.0

PR #620, opened 2018-03-15
Fixed SNAT examples.

ryan-talley:
Problem:
 Current SNAT examples were interleaved with other examples which could
 cause confusion to users.

Solution:
 Create separate SNAT example files to avoid the confusion.

Fixes: #619

### DIFF
--- a/docs/_static/config_examples/ingress-with-health-monitors.yaml
+++ b/docs/_static/config_examples/ingress-with-health-monitors.yaml
@@ -6,11 +6,6 @@ metadata:
   annotations:
     virtual-server.f5.com/ip:        "1.2.3.4"
     virtual-server.f5.com/partition: "k8s"
-    # Optionally specify the source address translation type of the virtual server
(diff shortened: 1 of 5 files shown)


sjberman:
Are there ci-docs that currently reference these?

ryan-talley:
No there is not the ci docs are currently in a draft state without references.

docs/_static/config_examples/example-vs-resource-source-address-translation.configmap.yaml
```diff
@@ -0,0 +1,35 @@
+kind: ConfigMap
```
sjberman: Just a nitpick, but I would shorten the name of this file since it's really long. Something like 'source-addr-translation.configmap.yaml'.

ryan-talley: sounds good will do.

PR #622, opened 2018-03-16
Moved JSON strings for configmaps out of appManager_test.go (fixes #597)

### LINKED ISSUE #597: Move unit test ConfigMap strings to actual json files
#### Description
Currently the appManager_test.go unit tests use JSON encoded strings for their configmaps which are difficult to edit. These should be moved into separate files json that can then be read in and parsed by the tests.

#### Controller Version
v1.4

dylan-way:
Problem:
The unit tests in file appManager_test.go made use of JSON encoded strings for
the configmap data. These strings were difficult to edit, and some of these
strings were unformatted.

Solution:
Each configmap has been formatted and moved from appManager_test.go into
its own JSON file. These files now reside in the pkg/appmanager/test folder
and are read in.

Fixes: #597

affects-branches: master


pkg/appmanager/appManager_test.go
```diff
@@ -44,280 +45,56 @@ func init() {
...
+var twoSvcTwoPodsConfig string = readConfigFile(configPath + "/twoSvcTwoPodsConfig_exp.json")
 
-var oneSvcTwoPodsConfig string = string(`{"resources":{"velcro":{"virtualServers":[{"name":"default_barmap","pool":"/velcro/cfgmap_default_barmap_bar","ipProtocol":"tcp","enabled":true,"sourceAddressTranslation":{"type":"automap"},"destination":"/velcro/10.128.10.240:6051","profiles":[{"partition":"Common","name":"http","context":"all"},{"partition":"Common","name":"tcp","context":"all"}]}],"pools":[{"name":"cfgmap_default_barmap_bar","loadBalancingMode":"round-robin","members":[{"address":"10.2.96.0","port":80,"session":"user-enabled"},{"address":"10.2.96.3","port":80,"session":"user-enabled"}]}]}}}`)
+var oneSvcTwoPodsConfig string = readConfigFile(configPath + "/oneSvcTwoPodsConfig_exp.json")
+
+func readConfigFile(path string) string {
+	data, err := ioutil.ReadFile(path)
+	// Returns emptyConfig is the file cannot be read
+	if nil != err {
+		return string(`{"resources":{}}`)
```
ryan-talley: Instead of returning an empty config here leverage ginkgo and use Expect(err).To(BeNil()) so that way we can catch the test failed since it tried to load a non-existent file rather than have the test failed because we ran with an empty config.

pkg/test/configmapBar.json
```diff
@@ -0,0 +1,20 @@
+{
```
ryan-talley: Move these to pkg/test/configs for better organization.

pkg/test/oneIappOneNodeConfig_exp.json
```diff
@@ -0,0 +1,97 @@
+{
```
ryan-talley: Follow go naming conventions so oneIappOneNodeConfig_exp -> oneIappOneNodeConfigExp

PR #631, opened 2018-03-22
Redid SNAT pool config.

ryan-talley:
Problem:
 Current per vs SNAT pool config did not provide the proper solution.

Solution
 Make a CLI configurable argument to specify a SNAT pool name that all
 virtual servers will reference if provided or config automap if not
 provided.


docs/README.rst
```diff
@@ -100,6 +100,12 @@ General
 |                       |         |          |                                  | to verify the BIG-IP                    |                |
 |                       |         |          |                                  | configuration.                          |                |
 +-----------------------+---------+----------+----------------------------------+-----------------------------------------+----------------+
+| vs-snat-pool-name     | string  | Optional | n/a                              | Name of the SNAT pool that all virtual  |                |
```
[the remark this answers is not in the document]
ryan-talley: Ya Booch said the they only care about automap or snat, since I guess it is difficult to set up the lists that are used when none is configured so we are leaving it out.

docs/README.rst
```diff
@@ -108,6 +114,10 @@ General
 
    :code:`python-basedir` optionally specifies the path to an alternate |kctlr| to F5 CCCL agent (:code:`bigipconfigdriver.py`). `F5 Controller Agent`_ is the default agent.
 
+   Use the ``vs-snat-pool-name`` if you want virtual servers to reference a preconfigured SNAT pool by name in the Common partition on the BIG-IP device.
```
russokj: So the user must create the snat pool in the Common partition, or can they create them in the controller's partition?

ryan-talley: They can create them in controller's partition since CCCL does not track SNAT pools, but we decided not to document this since we already recommend all preconfigured objects like health monitors be created in Common

russokj:
Looks like some conflicts exist.  Go ahead and merge after fixing.

ryan-talley:
sounds good will fix and merge. thanks for the reviews!

PR #637, opened 2018-03-28
Disallow single service Ingress to share VS

### LINKED ISSUE #608: Disallow single service Ingresses to share an IP address
#### Description
If a single-service Ingress exists for some IP address, a virtual server will be created, and its default pool will be set to that Ingress's service. If another single-service Ingress is created for the same IP address, the controller will attempt to share the virtual with this new Ingress, but it will still have the original default pool on the virtual. If this pool has gone away (due to Ingress1's service being deleted), this will cause the BIG-IP to error out trying to access a nonexistent pool (since the virtual still references it).

The solution here is to prevent single-service Ingresses from ever sharing an IP address. If an Ingress comes in and we already have a virtual for it, we should check if the virtual has a default pool. If it does, log an error that this IP address cannot be shared, and discontinue processing that Ingress.

This also makes sense to do because single service Ingresses do not have policies, so having different pools for one virtual, and no policy, doesn't really make sense.

#### Kubernetes Version
1.8

#### Controller Version
1.4

#### BIG-IP Version
12.1

sjberman:
Problem: Single service Ingresses should not be able to share a virtual server with any other Ingress. Single service Ingresses don't create policies (since there is only one pool), so having multiple pools without a policy doesn't make sense.

Solution: When an Ingress update happens, and either the incoming Ingress is single service, or we already have a single service configured on the specified IP:Port, then we reject the incoming Ingress.

Fixes #608

### DIFF
--- a/pkg/appmanager/appManager_test.go
+++ b/pkg/appmanager/appManager_test.go
@@ -2663,11 +2663,23 @@ var _ = Describe("AppManager Tests", func() {
 						f5VsPartitionAnnotation: "velcro",
 					})
 				// Add ingress with same ip (should use shared virtual)
-				ingress5 := test.NewIngress("ingressShared", "3", namespace,
+				ingress5 := test.NewIngress("ingressShared", "4", namespace,
 					v1beta1.IngressSpec{
-						Backend: &v1beta1.IngressBackend{
-							ServiceName: "foobar",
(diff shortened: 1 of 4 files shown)


russokj:
Looks good.  Fix the merge conflict and go ahead and merge the pull request.

PR #639, opened 2018-03-28
Fixed repeating INFO logs when a pool only config map or ingress is d…

### LINKED ISSUE #603: Repeated INFO logs when pool only config map deployed.
#### Description
When I deploy a pool only configmap I see this message in the container log files:
2018/03/07 16:49:18 [INFO] No virtual IP was specified for the virtual server k8s.poolonly creating pool only.
2018/03/07 16:49:27 [INFO] No virtual IP was specified for the virtual server k8s.poolonly creating pool only.
2018/03/07 16:49:27 [INFO] No virtual IP was specified for the virtual server k8s.poolonly creating pool only.
2018/03/07 16:49:33 [INFO] No virtual IP was specified for the virtual server k8s.poolonly creating pool only.
2018/03/07 16:49:33 [INFO] No virtual IP was specified for the virtual server k8s.poolonly creating pool only.
2018/03/07 16:49:35 [INFO] No virtual IP was specified for the virtual server k8s.poolonly creating pool only.
2018/03/07 16:49:35 [INFO] No virtual IP was specified for the virtual server k8s.poolonly creating pool only.

#### Kubernetes Version
<Version of Kubernetes being used>

#### Controller Version
CC 1.4.2

#### BIG-IP Version
13.0

#### Diagnostic Information

dylan-way:
…eployed. (fixes #603)

Problem:
When a pool only config map or ingress was deployed, the INFO log message
stating "No virtual IP was specified for the virtual server" would get
logged multiple times.

Solution:
This log message is now located in the validResources.go file within the
checkValidIngress and checkValidConfigMap methods rather than in the
resourceConfig.go file. There is now an additional check to guarantee that
the log message is only logged the first time a config is seen.

affects-branch: master

### DIFF
--- a/pkg/appmanager/validateResources.go
+++ b/pkg/appmanager/validateResources.go
@@ -17,6 +17,8 @@
 package appmanager
 
 import (
+	log "github.com/F5Networks/k8s-bigip-ctlr/pkg/vlogger"
+
 	routeapi "github.com/openshift/origin/pkg/route/api"
 	"k8s.io/client-go/pkg/api/v1"
 	"k8s.io/client-go/pkg/apis/extensions/v1beta1"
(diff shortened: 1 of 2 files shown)


pkg/appmanager/validateResources.go
```diff
@@ -42,6 +44,19 @@ func (appMgr *Manager) checkValidConfigMap(
...
+	// time we see a config.
+	rsName := formatConfigMapVSName(cm)
+	// Checking for annotation in VS, not iApp
+	if cfg.MetaData.ResourceType != "iapp" &&
+		cfg.Virtual.VirtualAddress != nil &&
+		cfg.Virtual.VirtualAddress.BindAddr == "" {
+		// Check for IP annotation provided by IPAM system
+		if _, ok := cm.ObjectMeta.Annotations[f5VsBindAddrAnnotation]; !ok {
+			log.Infof("No virtual IP was specified for the virtual server %s creating pool only.",
+				rsName)
```
sjberman: Don't forget to also verify that the config exists already, same as how you did it with Ingress.

pkg/appmanager/validateResources.go
```diff
@@ -138,7 +153,13 @@ func (appMgr *Manager) checkValidIngress(
 			}
 			return false, nil
 		}
-
+		// This ensures that pool-only mode only logs the message below the first
+		// time we see a config.
+		if _, exists := appMgr.resources.GetByName(rsName); !exists &&
+			bindAddr == "" {
```
sjberman: No need to have this on a newline here, you can just keep it with the line above (unless this line is abnormally long, in which case the log message below should be broken up into shorter lines).

PR #640, opened 2018-03-29
Whitelist Route default SSL profiles

### LINKED ISSUE #636: Supplying a default client/server SSL profile for a Route leads to an invalid configuration
#### Description
When using the k8s-bigip-ctlr to manage Route objects in Openshift, supplying a default client SSL profile for the resulting virtual server via the --default-client-ssl flag (or server SSL profile via --default-server-ssl) results in an invalid configuration, as the VS does not have any client profile applied to it. 

This can be verified by looking at the generated configuration in the controller container, and reproduced when supplying any SSL profile in this manner (ex /Common/clientssl fails, as does creating a profile inside the managed partition, ex /openshift/openshift-ssl, and supplying that)

Note that this is fairly silent - there is nothing in the log output that makes it plainly obvious that the controller is skipping over the supplied configuration items, it just does it, which leads to a route that listens for https traffic, but does not present a TLS certificate. Obviously, that's not quite ideal :)

Let me know if there's anything else you need me to provide on this.

Thanks!

#### Kubernetes Version
Openshift Origin 3.7 / kubernetes 1.7

#### Controller Version
1.4.2

#### BIG-IP Version
12.1.2

#### Diagnostic Information
I've attached the k8s-bigip-ctlr congiruation, the resultant pod terminal output with --log-level=DEBUG, and the generated configuration found in /tmp for both scenarios - supplying the --default-client-ssl flag and when it's absent. You can plainly see that when the flag is supplied, there is no matching configuration item in the generated configuration, but when the flag is absent, it generates the configuration without issue.

[k8s-bigip-ctlr-default.log](https://github.com/F5Networks/k8s-bigip-ctlr/files/1857394/k8s-bigip-ctlr-default.log)
[k8s-bigip-ctlr-profile-supplied.log](https://github.com/F5Networks/k8s-bigip-ctlr/files/1857395/k8s-bigip-ctlr-profile-supplied.log)
[k8s-bigip-ctlr-routes.yaml.txt](https://github.com/F5Networks/k8s-bigip-ctlr/files/1857396/k8s-bigip-ctlr-routes.yaml.txt)
[k8s-bigip-ctlr.config-default.json.txt](https://github.com/F5Networks/k8s-bigip-ctlr/files/1857397/k8s-bigip-ctlr.config-default.json.txt)
[k8s-bigip-ctlr.config-profile-supplied.json.txt](https://github.com/F5Networks/k8s-bigip-ctlr/files/1857398/k8s-bigip-ctlr.config-profile-supplied.json.txt)

sjberman:
Problem: When cleaning up profiles that no longer exist in a resource, we maintain a whitelist of profiles that don't come from any resource, so we know not to accidentally delete them. The client and server ssl profiles for Routes that are provided via CLI were not considered in this list.

Solution: Whitelist these two profiles to prevent them from being unnecessarily deleted. Also fixed an issue with the configuration of the server ssl profile name.

Fixes #636

### DIFF
--- a/pkg/appmanager/profiles.go
+++ b/pkg/appmanager/profiles.go
@@ -160,7 +160,7 @@ func (appMgr *Manager) setServerSslProfile(
 	}
 
 	// Handle the Default for SNI profile
-	appMgr.handleServerSNIDefaultProfile(rsCfg, peerCert)
+	appMgr.handleServerSNIDefaultProfile(rsCfg, peerCert, sKey)
 
 	if prof, ok := route.ObjectMeta.Annotations[f5ServerSslProfileAnnotation]; ok {
 		serverSsl, updated := appMgr.handleServerSslProfileAnnotation(
@@ -191,15 +191,13 @@ func (appMgr *Manager) setServerSslProfile(
 func (appMgr *Manager) handleServerSNIDefaultProfile(
 	rsCfg *ResourceConfig,
 	peerCert string,
+	sKey serviceQueueKey,
 ) {
 	if appMgr.routeConfig.ServerSSL != "" {
 		// User has provided a name
-		profile := ProfileRef{
-			Name:      appMgr.routeConfig.ServerSSL,
-			Partition: rsCfg.Virtual.Partition,
-			Context:   customProfileServer,
-		}
-		rsCfg.Virtual.AddOrUpdateProfile(profile)
+		prof := convertStringToProfileRef(
+			appMgr.routeConfig.ServerSSL, customProfileServer, sKey.Namespace)
+		rsCfg.Virtual.AddOrUpdateProfile(prof)
 	} else {
 		// No provided name, so we create a default
 		skey := secretKey{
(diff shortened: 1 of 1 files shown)


pkg/appmanager/profiles.go
```diff
@@ -191,15 +191,13 @@ func (appMgr *Manager) setServerSslProfile(
...
 	if appMgr.routeConfig.ServerSSL != "" {
 		// User has provided a name
-		profile := ProfileRef{
-			Name:      appMgr.routeConfig.ServerSSL,
-			Partition: rsCfg.Virtual.Partition,
-			Context:   customProfileServer,
-		}
-		rsCfg.Virtual.AddOrUpdateProfile(profile)
+		prof := convertStringToProfileRef(
+			appMgr.routeConfig.ServerSSL, customProfileServer, sKey.Namespace)
```
russokj: what happened to the Partition previously specified?  Is this handled internally by convertStringToProfileRef?

sjberman: It's handled internally by that function.

PR #641, opened 2018-03-30
Replace rule / with underscore

### LINKED ISSUE #638: path segments - e.g. `/foo/bar` - generate dead policy rule link in BigIP UI
#### Description
configuring an ingress with a spec of:
```
spec:
  rules:
  - http:
      paths:
      - backend:
          serviceName: svc-1-demo-web-app
          servicePort: 80
        path: /foo
      - backend:
          serviceName: svc-2-demo-web-app
          servicePort: 80
        path: /bar
      - backend:
          serviceName: svc-3-demo-web-app
          servicePort: 80
        path: /baz/qux
```
generates a working policy - traffic is able to reach the pools - however the link to the policy rule includes the latter `/`, resulting in a link within the BigIP UI that does not work and prevents viewing the rule conditions and actions. 

example URI:
`https://1.2.3.4/tmui/tmui/locallb/policy/app/#/policy/~test~ingress_1-2-3-4_80/rule/ingress__baz/qux_ingress_default_svc-3-demo-web-app?policy_status=legacy`

Where `ingress__baz/qux_ingress_default_svc-3-demo-web-app` should all be one segment, but is viewed by BigIP UI as two due to the underlying `baz/qux` multisegment path.

#### Kubernetes Version
1.8

#### Controller Version
From controller log:
`Starting: Version: 1.5.0-WIP, BuildInfo: n914-358525313`

#### BIG-IP Version
12.1.2

sjberman:
Problem: Extended paths that had / characters were causing dead links on the BIG-IP GUI. Traffic would still work, but the GUI would not.

Solution: Changed any / characters to underscores in Ingress rule names.

Fixes #638

### DIFF
--- a/pkg/appmanager/resourceConfig.go
+++ b/pkg/appmanager/resourceConfig.go
@@ -252,7 +252,9 @@ func formatIngressRuleName(host, path, pool string) string {
 	if path == "" {
 		rule = fmt.Sprintf("ingress_%s_%s", host, pool)
 	} else {
+		// Remove the first slash, then replace any subsequent slashes with '_'
 		path = strings.TrimPrefix(path, "/")
+		path = strings.Replace(path, "/", "_", -1)
 		rule = fmt.Sprintf("ingress_%s_%s_%s", host, path, pool)
 	}
 	return rule


recursivelycurious:
👍

PR #642, opened 2018-03-30
Fixes #539 

### LINKED ISSUE #539: [DOC] Further describe the behaviour of verify-interval
#### Description

After using k8s-bigip-ctrl I have found the description of this parameter could be further described, I would propose the following changes in [k8s-bigip-ctlr reference documentation](http://clouddocs.f5.com/products/connectors/k8s-bigip-ctlr/latest/)

- In the Danger frame: I would indicate that the periodicity of this config regeneration is done with the verify-interval. I find this is important because the term "verify" is a bit misleading given that the feature doesn't just verify the config but also enforces the expected configuration).
- I would indicate that the use of value 0 is discourage because at present it doesn't disable the verify-interval and that is the intended behaviour as per conversation with Matt Y and reported in https://github.com/F5Networks/k8s-bigip-ctlr/issues/538
- I would indicate that typically config regeneration time is very low (few ms) and don't cause service disruption but with very large configs this setting might be increased.

I hope this helps

#### Kubernetes Version
NA

#### Controller Version
All
#### BIG-IP Version
NA
#### Diagnostic Information
NA

jputrino:
- Fixes issue #539 
- light edits to further clarify/fix grammar for items in the general config parameters table

@alonsocamaro

### DIFF
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -68,7 +68,7 @@ Within the cluster, the allocated NodePort load balances traffic to all pods.
 
 .. danger::
 
-   The |kctlr| monitors the BIG-IP partition it manages for configuration changes. If it discovers changes, the Controller reapplies its own configuration to the BIG-IP system.
+   The |kctlr| monitors the BIG-IP partition it manages for configuration changes (see :code:`verify-interval` in the :ref:`General configuration parameters <general configs>` table). If it discovers changes, the Controller reapplies its own configuration to the BIG-IP system.
 
    F5 does not recommend making configuration changes to objects in any partition managed by the |kctlr| via any other means (for example, the configuration utility, TMOS, or by syncing configuration with another device or service group). Doing so may result in disruption of service or unexpected behavior.
 
@@ -78,7 +78,8 @@ Within the cluster, the allocated NodePort load balances traffic to all pods.
 
 Controller Configuration Parameters
 -----------------------------------
-All configuration parameters below are global to the |kctlr|.
+
+All of the configuration parameters below are global.
 
 .. _general configs:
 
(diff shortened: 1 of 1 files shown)


jputrino:
@f5yacobucci -- can you confirm what the `verify-interval` behavior is for iApps?

sjsharma2001:
I assigned this to @f5yacobucci.

f5yacobucci:
@jputrino The verify-interval does verify iapps and if they've changed or not. Are you asking that they are checked via the verify-interval or the specific methodology used and what's checked?

alonsocamaro:
@f5yacobucci @jputrino right now the CC re-instantiates the iApp every verify-interval period without checking if the args of the iApp has changed. Also the value of 0 is currently not documented.

jputrino:
@alonsocamaro I have documented the value of 0 here:
https://github.com/F5Networks/k8s-bigip-ctlr/pull/642/commits/f2a0f3c938dba11379deb08c23128c7cf4d713b5#diff-8ab5636c6c18641e5c58358ef489e0f7R130

alonsocamaro:
@jputrino I overlooked that. I still think the other clarification is useful.

jputrino:
@alonsocamaro - For the iApp, do you mean the CC re-instantiates without checking whether settings on the BIG-IP have changed? @f5yacobucci can speak to whether this is functioning by design or not.

alonsocamaro:
@jputrino Yes, that's the behaviour. And I checked with Matt. I think he has forgotten the conversation that we had and the behaviour change request that I submitted through an internal feature repository :-)

f5yacobucci:
I'm not sure if we should document the regeneration in this documentation. CCCL does attempt to to check iApps as it does other BIG-IP resources. It pulls down the existing iApps and uses the 'variables' and 'tables' sections explicitly to compare desired versus existing configurations.

If we're seeing iApps being overwritten too frequently this isn't a design choice and really a bug, the solution/product docs shouldn't be updated to document bugs - I think mentioning that behavior is more appropriately discussed in a "Known Bugs" or "Limitations" section.

alonsocamaro:
What I have found is that iApps get re-instantiated but that doesn't create new objects if these already exist. Creating new objects would cause service disruption. I'm afraid that when loading complex configurations such as ASM policies some disruption might occur but I haven't tested this as of yet.

jputrino:
@alonsocamaro I have opened a new issue for the investigation of the iApp verify-interval behavior (#665).

f5yacobucci:
Thanks @jputrino 
@alonsocamaro That is a valid concern. My only comment is that this isn't intentional on the controller's part; therefore, an oversight and bug. The permanent product documentation is the inappropriate place to discuss limitations due to known bugs. A document exists which discusses limitations, once #665 is investigated I'd suggest mentioning a limitation like this in the Release Notes, or better yet rectify it (another reason why we shouldn't document it in the product docs, once it's fixed it requires someone to remember to update these various statements).

PR #647, opened 2018-04-05
Added url-rewrite and app-root annotation processing. 

ryan-talley:


ryan-talley:
Currently a WIP since I still need to add the docs changes.

ryan-talley:
Removing WIP the docs changes are in.

pkg/appmanager/resourceConfig.go
```diff
@@ -1081,9 +1329,19 @@ func (appMgr *Manager) createRSConfigFromIngress(
...
+			appRootMap,
+			appMgr.sharedAnnotationRulesMap,
+			pools,
+			cfg.Virtual.Partition,
+		)
 		plcy = createPolicy(*rules, cfg.Virtual.Name, cfg.Virtual.Partition)
 	} else { // single-service
+		if len(appRootMap) != 0 {
+			log.Errorf("")
+		}
```
sjberman: Logging a blank error here.

pkg/appmanager/routing.go
```diff
@@ -182,13 +206,40 @@ func processIngressRules(
...
+					return nil, mergeRules, make(map[string]string), make(map[string][]string)
 				}
 				if true == strings.HasPrefix(uri, "*.") {
 					wildcards[uri] = rl
 				} else {
 					rlMap[uri] = rl
 				}
+
+				// Process url-rewrite annotation\
+				if !skipURLRewrite {
```
sjberman: Looks like this should be removed?

pkg/appmanager/resourceConfig.go
```diff
@@ -1353,10 +1675,14 @@ func (appMgr *Manager) createRSConfigFromRoute(
 					rsCfg.Pools[i].ServicePort = pool.ServicePort
 				}
 				found = true
+				// NOTE: should this be added?
+				break
```
sjberman: Yeah, this break is good here.

pkg/appmanager/resourceConfig.go
```diff
@@ -1440,6 +1774,16 @@ func (appMgr *Manager) handleRouteTls(
...
+				if len(appRootRules) == 2 {
+					rc.AddRuleToPolicy(policyName, appRootRules[0])
+					appMgr.AddRuleToAnnotationsMap(virtualName, poolName, appRootRules[0].Name)
+					rc.AddRuleToPolicy(policyName, appRootRules[1])
+					appMgr.AddRuleToAnnotationsMap(virtualName, poolName, appRootRules[1].Name)
+				}
+				if urlRewriteRule != nil {
+					rc.AddRuleToPolicy(policyName, urlRewriteRule)
+					appMgr.AddRuleToAnnotationsMap(virtualName, poolName, urlRewriteRule.Name)
+				}
```
sjberman: This block looks like it gets reused 4 times. Can we move it into its own small function?

pkg/appmanager/resourceConfig.go
```diff
@@ -1534,16 +2001,18 @@ func (rc *ResourceConfig) AddRuleToPolicy(
...
 ) {
 	// We currently have at most 1 policy, 'forwarding'
 	policy := rc.FindPolicy("forwarding")
 	if nil != policy {
 		for i, r := range policy.Rules {
 			if r.Name == rule.Name && r.FullURI == rule.FullURI {
 				// Remove old rule
+				unmerged := rc.UnmergeRule(rule.Name, mergedRulesMap)
 				if len(policy.Rules) == 1 {
 					rc.RemovePolicy(*policy)
```
sjberman: Only remove the policy if len() is 1 AND !unmerged. If the rule we are deleting was merged with another, and we unmerge and delete the initial rule, we are left with the unmerged rule (which shouldn't be deleted), but we will delete the policy anyway because there is 1 rule.

pkg/appmanager/resourceConfig.go
```diff
@@ -836,6 +838,234 @@ func setSourceAddrTranslation(snatPoolName string) SourceAddrTranslation {
...
+	var err error
+
+	if target != "" {
+		if !(strings.HasPrefix(target, "http://") || strings.HasPrefix(target, "https://")) {
+			target = "http://" + target
+		}
+
+		targetURL, err = url.Parse(target)
+		if err != nil {
+			log.Errorf("Error parsing url-rewrite target url: %s, Error: %v, skipping", target, err)
```
ryan-talley: Change to warning

pkg/appmanager/resourceConfig.go
```diff
@@ -836,6 +838,234 @@ func setSourceAddrTranslation(snatPoolName string) SourceAddrTranslation {
...
+				Path:            true,
+				Present:         true,
+				Remote:          true,
+				Request:         true,
+				Values:          []string{targetURL.Path},
+			})
+		}
+	}
+
+	if !(strings.HasPrefix(value, "http://") || strings.HasPrefix(value, "https://")) {
```
ryan-talley: Checkou routing.go to possibly combine conditions

pkg/appmanager/resourceConfig.go
ryan-talley: Change all Errors to warnings if they are recoverable.

pkg/appmanager/resourceConfig.go
```diff
@@ -1758,6 +2231,291 @@ func (rc *ResourceConfig) RemovePoolAt(offset int) bool {
...
+							mergeeEntry.RuleName = iName
+							mergerEntry.OtherRuleNames = []string{iName}
+							mergeeEntry.OtherRuleNames = []string{jName}
+							mergerEntry.OriginalActions = rules[j].Actions
+							mergeeEntry.OriginalActions = rules[i].Actions
+
+							// Merge only unique actions
+							for k := range rules[i].Actions {
+								found := false
+								for l := range rules[j].Actions {
```
ryan-talley: See if this can be functionalized

pkg/appmanager/resourceConfig.go
```diff
@@ -1758,6 +2231,291 @@ func (rc *ResourceConfig) RemovePoolAt(offset int) bool {
...
+				break
+			}
+		}
+
+		// Unmerge the rules that were merged with this rule
+		for _, mergeeRuleName := range entry.OtherRuleNames {
+			mergeeRuleEntry := mergedRulesMap[rsName][mergeeRuleName]
+			mergeeRule := &Rule{
+				Name:       mergeeRuleName,
+				Conditions: mergerConditions,
```
ryan-talley: Need to consider FullURI and Ordinal

pkg/appmanager/resourceConfig.go
```diff
@@ -1758,6 +2231,291 @@ func (rc *ResourceConfig) RemovePoolAt(offset int) bool {
...
+		for i := range policy.Rules {
+			if policy.Rules[i].Name == mergerRuleName {
+				var deletedActionIndices []int
+
+				// Find and these merged actions indices
+				for j := range mergedActions {
+					for k := range policy.Rules[i].Actions {
+						if mergedActions[j] == policy.Rules[i].Actions[k] {
+							deletedActionIndices = append(deletedActionIndices, k)
+							continue
```
ryan-talley: do not need this

pkg/appmanager/resourceConfig.go
ryan-talley: Removed

pkg/appmanager/routing.go
ryan-talley: Removed slash

pkg/appmanager/resourceConfig.go
ryan-talley: Removed comment left break

pkg/appmanager/resourceConfig.go
ryan-talley: moved

pkg/appmanager/resourceConfig.go
ryan-talley: Fixed

pkg/appmanager/resourceConfig.go
ryan-talley: Changed Errorf to Warningf

pkg/appmanager/routing.go
```diff
@@ -157,14 +157,38 @@ func createPolicy(rls Rules, policyName, partition string) *Policy {
...
+
+	var skipURLRewrite bool
+	if urlRewriteVal, ok := urlRewriteMap["single"]; ok == true {
+		urlRewriteRule := processURLRewrite(uri, urlRewriteVal)
+		urlRewriteRules = append(urlRewriteRules, urlRewriteRule)
+		skipURLRewrite = true
+	}
+
+	var skipAppRoot bool
+	if appRootVal, ok := appRootMap["string"]; ok == true {
```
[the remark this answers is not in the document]
ryan-talley: Fixed.

pkg/appmanager/resourceConfig.go
```diff
@@ -1043,10 +1273,28 @@ func (appMgr *Manager) createRSConfigFromIngress(
 	}
 	cfg.Virtual.Name = formatIngressVSName(bindAddr, pStruct.port)
 
+	// Handle url-rewrite annotation
+	var urlRewriteMap map[string]string
+	urlRewrite := ing.ObjectMeta.Annotations[f5VsURLRewriteAnnotation]
```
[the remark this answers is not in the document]
ryan-talley: Fixed

pkg/appmanager/routing.go
```diff
@@ -182,13 +206,40 @@ func processIngressRules(
 				rl, err = createRule(uri, poolName, partition, ruleName)
 				if nil != err {
 					log.Warningf("Error configuring rule: %v", err)
-					return nil
+					return nil, mergeRules, make(map[string]string), make(map[string][]string)
```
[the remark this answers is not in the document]
ryan-talley: changed

pkg/appmanager/routing.go
```diff
@@ -157,14 +157,38 @@ func createPolicy(rls Rules, policyName, partition string) *Policy {
...
+	var appRootRules []*Rule
+	var mergeRules bool
+
 	rlMap := make(ruleMap)
 	wildcards := make(ruleMap)
+	urlRewriteRefs := make(map[string]string)
+	appRootRefs := make(map[string][]string)
+
+	var skipURLRewrite bool
+	if urlRewriteVal, ok := urlRewriteMap["single"]; ok == true {
```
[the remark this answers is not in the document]
ryan-talley: changed

pkg/appmanager/resourceConfig.go
```diff
@@ -1337,6 +1629,36 @@ func (appMgr *Manager) createRSConfigFromRoute(
 		return &rsCfg, err, Pool{}
 	}
 
+	// Handle url-rewrite annotation
+	var urlRewriteRule *Rule
+	if urlRewrite, ok := route.ObjectMeta.Annotations[f5VsURLRewriteAnnotation]; ok == true {
```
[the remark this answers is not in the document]
ryan-talley: fixed

pkg/appmanager/resourceConfig.go
```diff
@@ -1498,10 +1862,113 @@ func (appMgr *Manager) handleRouteTls(
...
+					}
+					if urlRewriteRule != nil {
+						rc.AddRuleToPolicy(policyName, urlRewriteRule)
+						appMgr.AddRuleToAnnotationsMap(virtualName, poolName, urlRewriteRule.Name)
+					}
+				}
+			}
+		}
+	}
+	if urlRewriteRule != nil || len(appRootRules) != 0 {
```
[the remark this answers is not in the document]
ryan-talley: Fixed.

pkg/appmanager/resourceConfig.go
ryan-talley: removed

pkg/appmanager/resourceConfig.go
ryan-talley: looked at routing.go it uses multiple conditions, but I need to make path segments resolvable for these annotation rules.

docs/_static/config_examples/sample-url-rewrite-annotation.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: extensions/v1beta1
```
f5yacobucci: Is this the only url rewrite example? We should definitely have samples showing multi-path / multi-target examples as that's a complicated case.

pkg/appmanager/resourceConfig.go
```diff
@@ -836,6 +838,234 @@ func setSourceAddrTranslation(snatPoolName string) SourceAddrTranslation {
...
+	numSeps := strings.Count(annotation, ",")
+	numReps := strings.Count(annotation, "=")
+	if numSeps > 0 {
+		splits := strings.Split(annotation, ",")
+		for _, val := range splits {
+			if strings.Count(val, "=") != 1 {
+				log.Warningf("Annotation: %s value: %s not properly formatted should be replace-val=target-val, skipping", annotation, val)
+				continue
+			}
+			split := strings.Split(val, "=")
```
f5yacobucci: what happens if i make this mistake? "/hello=/good-bye/hi=/bye"?

it looks like i'll configured this: map["/hello"] = "/good-bye/hi" which is not really correct. plus i won't know why i don't get the map["/hi"] = "/bye" redirect.

we should do more checking around these splits.

docs/_static/config_examples/sample-url-rewrite-annotation.yaml
ryan-talley: There are better examples in the solution docs I can pull in I was thinking of having simple examples here and then the more complicated examples in the solution docs.

pkg/appmanager/resourceConfig.go
ryan-talley: This will fail since you have two equals signs thats the numSes == numReps check you do so even if you did "/hello=/good-bye/hi=/bye," it would still fail, but this could fail if you have ",/hello=/good-bye/hi=/bye,"  since num , == num =

pkg/appmanager/resourceConfig.go
ryan-talley: Tried doing this I am not coming up with anything quickly since it looks like rules are being passed by val rather than ref so due to time constraints I am going to keep this since it works.

pkg/appmanager/resourceConfig_test.go
```diff
@@ -937,4 +937,436 @@ var _ = Describe("Resource Config Tests", func() {
...
+			Expect(len(rule.Actions)).To(Equal(2))
+			Expect(rule.Actions[0].Name).To(Equal("0"))
+			Expect(rule.Actions[0].HTTPHost).To(Equal(true))
+			Expect(rule.Actions[0].Replace).To(Equal(true))
+			Expect(rule.Actions[0].Value).To(Equal("host.com"))
+			Expect(rule.Actions[1].Name).To(Equal("1"))
+			Expect(rule.Actions[1].HTTPURI).To(Equal(true))
+			Expect(rule.Actions[1].Path).To(Equal(""))
+			Expect(rule.Actions[1].Replace).To(Equal(true))
+			Expect(rule.Actions[1].Value).To(Equal("/path"))
```
sjberman: Seems a bit overkill to be testing every value in the actions for all of these test cases. I would say the only values we care about are Name and Value (for both actions and conditions). If those values equal what we expect, then we know all other values are "true", since they can't be false unless the entire rule is uninitialized.

pkg/appmanager/resourceConfig.go
```diff
@@ -1498,10 +1848,104 @@ func (appMgr *Manager) handleRouteTls(
...
+	if urlRewriteRule != nil || len(appRootRules) != 0 {
+		rc.MergeRules(appMgr.mergedRulesMap)
+	}
+}
+
+func (appMgr *Manager) AddRuleToAnnotationsMap(virtualName, poolName, ruleName string) {
+	if _, ok := appMgr.annotationRulesMap[virtualName]; !ok {
+		appMgr.annotationRulesMap[virtualName] = make(map[string][]string)
+	}
+	appMgr.annotationRulesMap[virtualName][poolName] = append(appMgr.annotationRulesMap[virtualName][poolName], ruleName)
```
sjberman: You should check if the rule you are about to add is already in the list. If it is, we don't want to add it again.

pkg/appmanager/resourceConfig_test.go
ryan-talley: I like being this explicit in this test since values are used for both HTTPURI and HTTPHost so if this function was written incorrectly you could get a rule of HTTPHost: true, Values: [/path] which is incorrect but the test would still pass since we only check name and values.

pkg/appmanager/resourceConfig.go
ryan-talley: Good catch I'll fix that

pkg/appmanager/resourceConfig.go
ryan-talley: Fixed.

pkg/appmanager/resourceConfig_test.go
```diff
@@ -1005,133 +1005,408 @@ var _ = Describe("Resource Config Tests", func() {
...
+				HTTPURI:         true,
+				Path:            true,
+				Present:         true,
+				Remote:          true,
+				Request:         true,
+				Values:          []string{"/approot"},
+			}}))
 		})
 
 		It("processes app-root annotations correctly", func() {
```
sjberman: Switch the test names to match (you flipped the test cases).

pkg/appmanager/resourceConfig_test.go
```diff
@@ -1005,133 +1005,408 @@ var _ = Describe("Resource Config Tests", func() {
...
+				},
+				&action{
+					Name:    "1",
+					HTTPURI: true,
+					Path:    "/path",
+					Replace: true,
+					Request: true,
+					Value:   "/newpath",
+				},
+			}))
```
sjberman: You should do one test case that uses a multi-segmented path, to be sure that you get all the conditions for each segment.

ryan-talley: Added.

PR #648, opened 2018-04-05
Don't remove Route annotation profiles

### LINKED ISSUE #649: Route annotations profiles get deleted
#### Description
If profiles are provided via annotations to Routes, the controller deletes them before the configs are written out to the BIG-IP. This shouldn't be happening. These profiles need to be verified that they are in use so the controller doesn't delete them.

#### Kubernetes Version
1.8

#### Controller Version
1.4

#### BIG-IP Version
12.1

sjberman:
Problem: If pre-existing profiles were being supplied via annotation for Routes, the controller was accidentally deleting these profiles.

Solution: When cleaning up unused profiles, be sure to check if the annotation profiles should be kept. The client/server profiles we check will either be in the route OR in the annotation.

Fixes #649

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -24,6 +24,7 @@ Bug Fixes
 * :issues:`608` - Single service Ingresses cannot share virtual servers.
 * :issues:`636` - Controller configures default ssl profiles for Routes when specified via CLI.
 * :issues:`638` - Ingress extended paths no longer break BIG-IP GUI links.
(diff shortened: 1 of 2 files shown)


amudukutore:
Is this an issue we have with the current released version of the product? If so, can you please file an issue for it so we can add it to the release notes.

sjberman:
@amudukutore Bug and release note added to this PR.

amudukutore:
@sjberman - thanks. The release notes and bug look good to me. Will defer to @russokj on code changes.

PR #658, opened 2018-04-11
Fix multi-service to single-service transition

sjberman:
Problem: When updating a multi-service Ingress to a single-service Ingress, there were a few issues. First, the controller would still have the previous multi-service ingress stored, and so when seeing the single-service, would deny the IP address to be shared, and reject the Ingress. Also, if changing from a multi-service to single-service, the default pool was not updated.

Solution: Change the single-service check to verify whether or not the previous config came from an ingress with the same name, rather than of the same type. This allows different type ingresses to be updated properly. If going from a multi-service Ingress to single-service, set the virtual's default pool.

### DIFF
--- a/pkg/appmanager/resourceConfig.go
+++ b/pkg/appmanager/resourceConfig.go
@@ -1047,6 +1047,7 @@ func (appMgr *Manager) createRSConfigFromIngress(
 	var pools Pools
 	var plcy *Policy
 	var rules *Rules
+	var ssPoolName string
 	if nil != ing.Spec.Rules { //multi-service
 		for _, rule := range ing.Spec.Rules {
 			if nil != rule.IngressRuleValue.HTTP {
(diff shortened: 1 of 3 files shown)


pkg/appmanager/resourceConfig.go
```diff
@@ -1094,10 +1095,11 @@ func (appMgr *Manager) createRSConfigFromIngress(
 			ServiceName: ing.Spec.Backend.ServiceName,
 			ServicePort: ing.Spec.Backend.ServicePort.IntVal,
 		}
+		ssPoolName = pool.Name
```
russokj: I'm not sure I see the significance of this change.  Is this just cosmetic?

russokj: Never mind.  The function keeps going.  I see where it's used now.

PR #663, opened 2018-04-16
Removed GUIDs from rule names

ryan-talley:
Problem:
 Rule names with GUIDS were causing rules to be reprocessed since a rule
 created when the config is compared with the old config will have
 differing names for the same annotations since they are GUIDs.

Solution:
 Prevent reprocessing by setting the rule name to the annotation target
 and value values which is unique, but will result in the same name for
 rules created from the same annotation.


sjberman:
Actually, one request. In both createRSConfig functions where ProcessAnnotations is called, could you put in a comment that says something like:

"This is called to ensure that if a url-rewrite annotation is changed for a pool, the old value is cleaned up. All rules will be unmerged for the pool, but those that still remain will be re-merged before writing out the final config. Only deleted annotations will be completely removed."

ryan-talley:
Sure I'll add those comments. Also I am going to merge this after removing the changes in the dockerfiles those are going to be addressed separately.

PR #670, opened 2018-04-18
Add hooks to do strict license checking when building on stable branches/releases

sjsharma2001:


.travis.yml
```diff
@@ -33,11 +33,14 @@ script:
   - |
     if [[ "$TRAVIS_BRANCH" == *"-stable" ]]; then
       export CTLR_VERSION=v$(echo $TRAVIS_BRANCH | sed s/-stable//g)
+      export LICENSE_STRICT="true"
```
amudukutore: I recall we had a discussion on this on an email thread (or perhaps on Slack) as to when this should be enabled. If I recall, we should only enable strict checking on tagged release builds.

Makefile
```diff
@@ -136,6 +136,9 @@ pip_attributions.json: always-build
 		--project-path=$(CURDIR) \
 
 docs/_static/ATTRIBUTIONS.md: flatfile_attributions.json  golang_attributions.json  pip_attributions.json
+        ifeq ($(LICENSE_STRICT), "true")
```
amudukutore: What happens to this check if LICENSE_STRICT is not defined?

.travis.yml
sjsharma2001: I have pushed that change.

sjberman:
@amudukutore I suggested to Sandeep to create his own *-stable branch in his fork, and push the changes to that and ensure the build runs correctly.

amudukutore:
@sjberman - that's a great idea. He can then create a tag on the stable branch on his fork and test the strict=true behavior.

sjsharma2001:
Yes, I will do these tests before merging..

sjsharma2001:
I tested the cases of pushing to "*-stable" branch, as well as creating a tag. Had to make few edits to make it work correctly.

Makefile
```diff
@@ -11,9 +11,23 @@ export BUILD_INFO := $(shell ./build-tools/version-tool build-info)
 
 GO_BUILD_FLAGS=-v -ldflags "-extldflags \"-static\" -X main.version=$(BUILD_VERSION) -X main.buildInfo=$(BUILD_INFO)"
 
-# Allow users to pass in BASE_OS build options (alpine or rhel7)
+# Allow users to pass in BASE_OS build options (alpine or rhel7)
```
sjberman: Remove the trailing space here.

Makefile
```diff
@@ -135,7 +149,8 @@ pip_attributions.json: always-build
 		--requirements=requirements.txt \
 		--project-path=$(CURDIR) \
 
+
```
sjberman: Remove this extra line.

Makefile
```diff
@@ -11,9 +11,23 @@ export BUILD_INFO := $(shell ./build-tools/version-tool build-info)
...
 GO_BUILD_FLAGS=-v -ldflags "-extldflags \"-static\" -X main.version=$(BUILD_VERSION) -X main.buildInfo=$(BUILD_INFO)"
 
-# Allow users to pass in BASE_OS build options (alpine or rhel7)
+# Allow users to pass in BASE_OS build options (alpine or rhel7) 
 BASE_OS ?= alpine
 
+# This is for builds not triggered through Travis CI 
+ifndef LICENSE_STRICT
+	LICENSE_STRICT=false
+endif
```
sjberman: I think there is a simpler way to do this. `LICENSE_STRICT ?= false` should set the variable to false if it isn't already set.

Makefile
```diff
@@ -11,9 +11,23 @@ export BUILD_INFO := $(shell ./build-tools/version-tool build-info)
...
+
+# Flag for level of license approval check 
+# (default if no flag is passed is devel)
+LIC_FLAG=
```
sjberman: Is this actually doing anything?

sjsharma2001: I am going to try  with this line removed and see if things work as expected.

Makefile
sjsharma2001: Will do.

Makefile
sjsharma2001: Will do.

Makefile
sjsharma2001: Per our discussion, if testing is taking too long to verify all scenarios, I will defer it for now.

sjsharma2001:
Hi @sjberman I have made all the changes you recommended, and they worked fine in my testing.

PR #672, opened 2018-04-18
Annotation error logging workaround.

ryan-talley:
Problem:
 Error logging was printing a lot because of the large amount of
 processing in the controller. This was causing logs of incorrect
 annotations to spam the log.

Solution:
 Move the logging and annotation validation into validateResources
 functions which are only called when the resource update is first
 recieved. This prevents multiple print outs of the same log message.


pkg/appmanager/validateResources.go
```diff
@@ -219,3 +276,60 @@ func (appMgr *Manager) checkValidRoute(
...
+		targetURL := parseAnnotationURL(k)
+		valueURL := parseAnnotationURL(v)
+
+		if rsType == multiServiceIngressType && targetURL.Host == "" {
+			log.Warningf("Invalid annotation: %s=%s need a host target for app-root annotation for multi-service ingress, skipping", k, v)
+			return
+		}
+		if rsType == multiServiceIngressType && targetURL.Path != "" {
+			if targetURL.Path != valueURL.Path {
+				return
```
sjberman: This case isn't really doing anything.

pkg/appmanager/validateResources.go
```diff
@@ -155,6 +158,34 @@ func (appMgr *Manager) checkValidIngress(
...
+		if appRoot, ok := ing.ObjectMeta.Annotations[f5VsAppRootAnnotation]; ok {
+			appRootMap := parseAppRootURLRewriteAnnotations(appRoot)
+			if rsType == singleServiceIngressType {
+				if len(appRootMap) > 1 {
+					log.Warning("Single service ingress does not support multiple url-rewrite annotation values, not processing")
+				} else {
+					if _, ok := appRootMap["single"]; ok {
+						validateAppRootAnnotations(rsType, appRootMap)
+					} else {
+						log.Warningf("url-rewrite annotation: %s has targeted values which is not supported for single service ingress, not processing", appRoot)
```
sjberman: Wrong warning here, should be an app-root warning.

pkg/appmanager/validateResources.go
```diff
@@ -155,6 +158,34 @@ func (appMgr *Manager) checkValidIngress(
...
+				log.Warning("Single service ingress does not support url-rewrite annotation, not processing")
+			}
+		}
+
+		// Validate app-root annotations
+		if appRoot, ok := ing.ObjectMeta.Annotations[f5VsAppRootAnnotation]; ok {
+			appRootMap := parseAppRootURLRewriteAnnotations(appRoot)
+			if rsType == singleServiceIngressType {
+				if len(appRootMap) > 1 {
+					log.Warning("Single service ingress does not support multiple url-rewrite annotation values, not processing")
```
sjberman: Wrong warning here, should be app-root.

pkg/appmanager/resourceConfig.go
```diff
@@ -1664,13 +1648,9 @@ func (appMgr *Manager) createRSConfigFromRoute(
 	var urlRewriteRule *Rule
 	if urlRewrite, ok := route.ObjectMeta.Annotations[f5VsURLRewriteAnnotation]; ok {
 		urlRewriteMap := parseAppRootURLRewriteAnnotations(urlRewrite)
-		if len(urlRewriteMap) > 1 {
-			log.Warningf("URL rewrite annotation: %s does not support multiple values for routes, not processing", urlRewrite)
```
sjberman: I don't see this warning moved anywhere, we still want to use it though, right?

pkg/appmanager/resourceConfig.go
```diff
@@ -1679,13 +1659,9 @@ func (appMgr *Manager) createRSConfigFromRoute(
 	var appRootRules []*Rule
 	if appRoot, ok := route.ObjectMeta.Annotations[f5VsAppRootAnnotation]; ok {
 		appRootMap := parseAppRootURLRewriteAnnotations(appRoot)
-		if len(appRootMap) > 1 {
-			log.Warningf("App root annotation: %s does not support multiple values for routes, not processing", appRoot)
```
sjberman: Same comment above about this warning not being moved somewhere else.

pkg/appmanager/validateResources.go
ryan-talley: Removed.

pkg/appmanager/validateResources.go
ryan-talley: Fixed.

pkg/appmanager/validateResources.go
ryan-talley: Fixed.

pkg/appmanager/resourceConfig.go
ryan-talley: This moved to validateResources line 253

pkg/appmanager/resourceConfig.go
ryan-talley: This moved to validateResources line 265

PR #676, opened 2018-04-20
Fix url-rewrite rule naming add limitations to release notes.

ryan-talley:
Problem:
 url-rewrite rule names were targetvalue instead of target-value making
 them difficult to read. The release notes were missing the limitation
 comment for the url-rewrite feature.

Solution:
 Add - to url-rewrite names so they are target-value. Add limitation
 note to release notes.

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -37,6 +37,10 @@ Bug Fixes
 * :issues:`649` - Route annotation profiles are no longer ignored.
 * :cccl-issue:`214` - Keys and certificates are now installed onto the managed partition.
 
+Limitations
+```````````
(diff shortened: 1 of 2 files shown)


docs/RELEASE-NOTES.rst
```diff
@@ -37,6 +37,10 @@ Bug Fixes
 * :issues:`649` - Route annotation profiles are no longer ignored.
 * :cccl-issue:`214` - Keys and certificates are now installed onto the managed partition.
 
+Limitations
+```````````
+* Can not apply app-root and url-rewrite annotations to the same resource see: https://github.com/F5Networks/k8s-bigip-ctlr/issues/675
```
sjberman: Nit: Can not should be cannot, and I would put a semicolon between "resource; see".

PR #711, opened 2018-06-22
Add type field to virtual-server.f5.com/health Ingress annotation

mzahorik:
Problem: The virtual-server.f5.com/serverssl Ingress annotation was
recently introduced, but the backend health monitors were hardcoded
to use http checks. Pool members using https would fail checks when
http requests were sent.

Solution: Add the new optional field "type" in
virtual-server.f5.com/health. This allows users to specify https as
a monitor type.


pkg/appmanager/healthMonitors.go
```diff
@@ -80,14 +80,18 @@ func (appMgr *Manager) assignMonitorToPool(
...
+			monitorType := ruleData.healthMon.Type
+			if monitorType == "" {
+				monitorType = "http"
+			}
 			monitor := Monitor{
 				// Append the protocol to the monitor names to differentiate them.
 				// Also add a monitor index to the name to be consistent with the
 				// marathon-bigip-ctlr. Since the monitor names are already unique here,
 				// appending a '0' is sufficient.
 				Name:      formatMonitorName(poolName),
```
mzahorik: I'll note I didn't mess with the formatMonitorName function.  It appends the suffix _0_http to the monitor, which can be confusing if it's an https monitor.  I thought to remove the _http completely, but wanted to verify that this was safe before doing so.  If the _http suffix stays, it's just cosmetic, the monitor correctly shows its type as https.

amudukutore: @mzahorik  - Thanks for the contribution. Can you please make sure you follow the guidelines in https://github.com/F5Networks/k8s-bigip-ctlr/blob/master/CONTRIBUTING.md, specifically, the section on signing the CLA: https://github.com/F5Networks/k8s-bigip-ctlr/blob/master/CONTRIBUTING.md#contributor-license-agreement? Thanks.

amudukutore: @mzahorik - we want to fix this problem. If you haven't signed the CLA, can you please do so? There is some urgency on this since we want to get a release out in the next week or so. If we don't hear back from you, we'll make this fix and also look at fixing the _http suffix along with it.

PR #716, opened 2018-06-27
Add --pull flag to docker build command

sjsharma2001:


amudukutore:
Have you tested with both rhel7 and non-rhel7 builds?

build-tools/build-devel-image.sh
```diff
@@ -39,7 +43,7 @@ cp $CURDIR/entrypoint.builder.debian.sh $WKDIR
 #cp $CURDIR/../python/k8s-*-requirements.txt $WKDIR/
 #cp $CURDIR/../requirements.docs.txt $WKDIR
 
-docker build --force-rm ${NO_CACHE_ARGS} \
+docker build $PULL_FLAG --force-rm ${NO_CACHE_ARGS} \
```
amudukutore: What happens here when BASE_OS is not "rhel7"? Does PULL_FLAG default to empty?

sjsharma2001: Yes, it defaults to empty in that case.
I have tested it for both rhel7 and non-rhel7 cases and the code behaves correctly. (--pull for rhel7 only)

PR #723, opened 2018-07-19
Fix performance issues

sjberman:
Problem: There were a couple of problems causing performance issues. First, using gob encoding to deepcopy configs proved to be inefficient. Secondly, the controller's steady-state check relied upon the queue being empty before writing any configs. With a large number of resources to process, and Kubernetes constantly updating endpoints, the queue would take a long time before being empty, therefore the controller wouldn't write any configs in a reasonable time.

Solution: Switched back to manual deepcopy instead of gob encoding. Saved off the initial queue length on startup, and kept track of the number of processed items before writing out a config, rather than waiting for the queue to be empty.

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -79,6 +79,8 @@ type Manager struct {
 	routeClientV1     rest.Interface
 	configWriter      writer.Writer
 	initialState      bool
+	queueLen          int
+	processedItems    int
 	// Use internal node IPs
 	useNodeInternal bool
 	// Running in nodeport (or cluster) mode
@@ -175,6 +177,8 @@ func NewManager(params *Params) *Manager {
 		useNodeInternal:    params.UseNodeInternal,
 		isNodePort:         params.IsNodePort,
 		initialState:       params.initialState,
+		queueLen:           0,
+		processedItems:     0,
 		routeConfig:        params.RouteConfig,
 		nodeLabelSelector:  params.NodeLabelSelector,
 		resolveIng:         params.ResolveIngress,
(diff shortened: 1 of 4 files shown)


amudukutore:
@sjberman - could you split the changes into 2 commits - one for deepcopy and another one for the queue-length changes? Thanks.

pkg/appmanager/resourceConfig.go
```diff
@@ -1742,16 +1740,41 @@ func (appMgr *Manager) createRSConfigFromRoute(
...
-	err = gob.NewDecoder(&cfgBytes).Decode(&rc)
-	if err != nil {
-		log.Errorf("Could not decode ResourceConfig: %v", err)
-		return
+	// MetaData
+	rc.MetaData = cfg.MetaData
+	// Virtual
+	rc.Virtual = cfg.Virtual
+	// Profiles
+	rc.Virtual.Profiles = make([]ProfileRef, len(cfg.Virtual.Profiles))
```
russokj: I don't see rc.Virtual.Policies or rc.Virtual.IRules being copied over like you are doing for rc.Virtual.Profiles.

pkg/appmanager/resourceConfig.go
```diff
@@ -1742,16 +1740,41 @@ func (appMgr *Manager) createRSConfigFromRoute(
...
+	copy(rc.Virtual.Profiles, cfg.Virtual.Profiles)
+	// Pools
+	rc.Pools = make(Pools, len(cfg.Pools))
+	copy(rc.Pools, cfg.Pools)
+	// Pool Members
+	for i, _ := range rc.Pools {
+		rc.Pools[i].Members = make([]Member, len(cfg.Pools[i].Members))
+		copy(rc.Pools[i].Members, cfg.Pools[i].Members)
```
russokj: Same thing for rc.Pools[i].MonitorNames (do we need to do a make/copy for it since it's an array?)

pkg/appmanager/appManager.go
```diff
@@ -753,6 +760,7 @@ func (appMgr *Manager) processNextVirtualServer() bool {
 
 	err := appMgr.syncVirtualServer(key.(serviceQueueKey))
 	if err == nil {
+		appMgr.processedItems++
```
russokj: Do we want to continue this forever (if ints are 64-bit we are okay, if they are 32-bit can't we eventually overflow?

russokj:
Minor comments.  Address if you need to and go ahead and self-merge.

PR #725, opened 2018-07-23
Allow files for BIG-IP parameters

sjberman:
Problem: Using the CLI to provide BIG-IP arguments is a security vulnerability, as they are displayed when running 'ps'.

Solution: Allow for users to create and mount files via a Secret, and provide the mount path to the controller. The controller will look in this path for username, password, and url files. If they are not found, it falls back to using the CLI.

### DIFF
--- a/cmd/k8s-bigip-ctlr/main.go
+++ b/cmd/k8s-bigip-ctlr/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
@@ -102,6 +103,7 @@ var (
 	bigIPUsername   *string
 	bigIPPassword   *string
 	bigIPPartitions *[]string
+	credsDir        *string
 
 	vxlanMode        string
 	openshiftSDNName *string
@@ -166,6 +168,9 @@ func _init() {
 		"Required, password for the Big-IP user account.")
 	bigIPPartitions = bigIPFlags.StringArray("bigip-partition", []string{},
 		"Required, partition(s) for the Big-IP kubernetes objects.")
+	credsDir = bigIPFlags.String("credentials-directory", "",
+		"Optional, directory that contains the BIG-IP username, password, and/or "+
+			"url files. To be used instead of username, password, and/or url arguments.")
 
 	bigIPFlags.Usage = func() {
 		fmt.Fprintf(os.Stderr, "  BigIP:\n%s\n", bigIPFlags.FlagUsagesWrapped(width))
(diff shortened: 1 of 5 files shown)


cmd/k8s-bigip-ctlr/main.go
```diff
@@ -369,6 +359,77 @@ func verifyArgs() error {
 	return nil
 }
 
+func getCredentials() error {
+	if len(*credsDir) > 0 {
+		var usr, pass, bigipURL string
+		var usrBytes, passBytes, urlBytes []byte
+		var err error
+		if strings.HasSuffix(*credsDir, "/") {
+			usr = *credsDir + "username"
```
f5yacobucci: This won't work with the .spec.volumes[].secret.items key and path functionality. If they place the files deeper into the directory structure or change the name we'll fail. I'm not suggesting we support this but we should call this out as a warning in the docs to NOT do that.

cmd/k8s-bigip-ctlr/main_test.go
```diff
@@ -324,6 +314,43 @@ var _ = Describe("Main Tests", func() {
 			Expect(err).ToNot(BeNil())
 		})
 
+		It("gets credentials from a file", func() {
```
f5yacobucci: We should add an additional test that verifies credentials-directory values supersede the other CLI arguments. Each one should be provided via both methods and make sure that the ones set come from the k8s secrets.

docs/README.rst
```diff
@@ -153,6 +153,23 @@ BIG-IP system
...
+     The :code:`credentials-directory` option is an alternative to using the :code:`bigip-username`, :code:`bigip-password`, or
+     :code:`bigip-url` arguments.
+
+     When you use this argument, the controller looks for three files in the specified directory:
+
+     - "username", "password", and "url"
+
+     If any of these files do not exist, the controller falls back to using the CLI arguments as parameters.
+
+     Each file should contain **only** the username, password, and url, respectively. You can create and mount
```
f5yacobucci: The above comment should be noted here, to *not* use the secret.items functionality.

cmd/k8s-bigip-ctlr/main.go
```diff
@@ -369,6 +359,77 @@ func verifyArgs() error {
...
+			pass = *credsDir + "password"
+			bigipURL = *credsDir + "url"
+		} else {
+			usr = *credsDir + "/username"
+			pass = *credsDir + "/password"
+			bigipURL = *credsDir + "/url"
+		}
+
+		usrBytes, err = ioutil.ReadFile(usr)
```
f5yacobucci: Just an aside, this could be an anonymous function that takes a string and pointer called three times.

cmd/k8s-bigip-ctlr/main.go
```diff
@@ -374,34 +373,32 @@ func getCredentials() error {
...
 		}
 
-		usrBytes, err = ioutil.ReadFile(usr)
-		if err != nil {
-			log.Debug("No username in credentials directory, falling back to CLI argument")
-			if len(*bigIPUsername) == 0 {
-				return fmt.Errorf("BIG-IP username not specified")
+		setField := func(field *string, filename, fieldType string) error {
+			var fileBytes []byte
+			fileBytes, err = ioutil.ReadFile(filename)
```
f5yacobucci: Do you mean to shadow err here? Because of the closure scope err is going to be the memory from outside this function. Might be safer to use := here to make sure we're using an err scoped to this function.

sjberman: go vet won't let me

cmd/k8s-bigip-ctlr/main.go
```diff
@@ -374,34 +373,32 @@ func getCredentials() error {
...
+			if err != nil {
+				log.Debug(fmt.Sprintf(
+					"No %s in credentials directory, falling back to CLI argument", fieldType))
+				if len(*field) == 0 {
+					return fmt.Errorf(fmt.Sprintf("BIG-IP %s not specified", fieldType))
+				}
+			} else {
+				*field = string(fileBytes)
```
f5yacobucci: Isn't a return nil required here?

f5yacobucci: n/m - diff confusion

PR #727, opened 2018-07-26
Config example for credentials directory

sjberman:
Adding deployment config example for how to mount a secret containing BIG-IP information into the controller.


amudukutore:
Please make the changes discussed offline. Thanks.

PR #736, opened 2018-08-23
Support LoadBalancer Service types in addition to NodePort.

maisem:
LoadBalancer and NodePort services behave the same way in the sense that they use a port on the node for traffic ingress. The controller should be able to treat them the same way.

I've tested that this works for NodePort and LoadBalancer service types.

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -1527,7 +1527,7 @@ func (appMgr *Manager) updatePoolMembersForNodePort(
 	rsCfg *ResourceConfig,
 	index int,
 ) (bool, string, string) {
-	if svc.Spec.Type == v1.ServiceTypeNodePort {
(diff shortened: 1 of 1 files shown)


karan:
@sjberman - Hi Saylor. Could you please take a look at this PR, or assign it to someone else who could. Greatly appreciate it.

Thanks

f5yacobucci:
Currently we have plans to more completely support type LoadBalancer Services beyond a drop in replacement for NodePort. There are semantic and functional differences between the two types outside of the Node level networking and we see LoadBalancer as a separate yet parallel configuration mechanism that would possibly function quite differently. 

What problem is this MR solving outside of the type support?

ssurenr:
Hi @maisem Could you us send a signed CLA? Guidelines are in https://github.com/F5Networks/k8s-bigip-ctlr/blob/master/CONTRIBUTING.md#contributor-license-agreement

PR #737, opened 2018-08-24
Adds switch to disable ConfigMap processing

caphrim007:
This patch adds a switch to the creation args to instruct CC to
either watch, or not watch, ConfigMap events. The default value
of this switch is `true`; as-in "watch ConfigMaps". This was
the default behavior before this switch was added.


amudukutore:
@caphrim007 - I am adding @sjberman as a reviewer. You will need to run our system tests against this change to make sure the default behavior is unchanged. If you need help getting these tests to run in our environment, I can help.

caphrim007:
@amudukutore I already have

PR #752, opened 2018-09-18
Adds whitelist annotation support to CC

caphrim007:
This patch adds whitelist annotation processing to CC. Whitelist
entries are created as part of L7 policies.

### DIFF
--- a/pkg/appmanager/routing.go
+++ b/pkg/appmanager/routing.go
@@ -186,6 +186,24 @@ func createPolicy(rls Rules, policyName, partition string) *Policy {
 
 	plcy.Rules = rls
 
+	// Check for the existence of the TCP field in the conditions.
+	// This would indicate that a whitelist rule is in the policy
+	// and that we need to add the "tcp" requirement to the policy.
+	requiresTcp := false
+	for _, x := range rls {
+		for _, c := range x.Conditions {
+			if c.Tcp == true {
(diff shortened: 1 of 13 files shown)


pkg/appmanager/routing.go
```diff
@@ -19,17 +19,16 @@ package appmanager
...
 	"sort"
 	"strconv"
 	"strings"
-	"sync"
-
-	log "github.com/F5Networks/k8s-bigip-ctlr/pkg/vlogger"
 
 	routeapi "github.com/openshift/origin/pkg/route/api"
 	"k8s.io/client-go/pkg/apis/extensions/v1beta1"
+	"sync"
```
sjberman: Leave the import statements as they were. Standard go convention for import structure is:

1. standard library
2. third party
3. internal packages

sjberman:
Also, please add this feature to the release notes for the next release.

If you could run the functional tests to verify nothing else was broken, that would be great.

caphrim007:
@sjberman ready for review.

sjberman:
Could you add a small test case(s) to the "merges and sorts rules correctly with matching conditions" unit test in resourceConfig_test.go to verify merging/unmerging of this new rule with existing rules works properly?

caphrim007:
@sjberman ready for review

Issue #735, opened 2018-08-23
k8s-bigip-ctlr fails to remove policy after the deletion of ingress


amudukutore:
#### Description
PROBLEM:  The customer created two ingresses that share the same IP, port, and service but with different rules. They noticed that when the second ingress is deleted the rules are not removed from the BIGIP.
**Note:** If there are two ingresses that share the same ip and port but different services, the controller properly removes the L7 traffic policy rules associated with an ingress when it is deleted.

IMPACT - The rules are not removed from a currently active policy on the BIGIP which may cause traffic interruption.

#### Kubernetes Version
1.11

#### Controller Version
v1.6.0

#### BIG-IP Version
13.1.0.7

#### Diagnostic Information
<Configuration files, error messages, logs>
Note: Sanitize the data. For example, be mindful of IPs, ports, application names and URLs
Note: The following F5 article outlines the information required when opening an issue.
https://support.f5.com/csp/article/K60974137

sjberman:
This likely impacts OpenShift Routes as well.

amudukutore:
This is most likely related to #753.

PR #756, opened 2018-09-27
Fix same service rule cleanup

ryan-talley:
Problem:
 Currently if there are two multi-service ingresses or routes pointed
 at the same service if one of them is deleted its rules will be left
 on the policy on the BIG-IP causing disruption to traffic as the
 controller tries to reconcile the desired config versus the left
 overs on the BIG-IP.

Solution:
 Add rule check in the dependency deletion in UpdateDependencies that
 will find the rules associated with cleaned up dependencies and add
 those to the remove dependencies slice to clean them up. Update unit
 tests to factor in this change.

Fixes: #735

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -9,6 +9,10 @@ Added Functionality
 * Added `--manage-configmaps` argument to CC to prevent or allow CC to respond to ConfigMap events. Defaults to `true`.
 * Added `virtual-server.f5.com/whitelist-source-range` annotation to support CIDR whitelisting.
 
+Bug Fixes
+`````````
+* :issues:`735` - Deleted rules from routes and ingresses on the same service not cleaned up properly.
+
 v1.6.1
 ------
 
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -1256,6 +1256,11 @@ func (appMgr *Manager) syncRoutes(
 						for _, rl := range pol.Rules {
 							if rl.FullURI == dep.Name {
 								rsCfg.DeleteRuleFromPolicy(pol.Name, rl, appMgr.mergedRulesMap)
+								// Delete profile (route only)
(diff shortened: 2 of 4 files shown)


sjberman:
Can you add a release note section for 1.6.2 and add this bug fix to it? The release notes for the "next-release" should be left alone (those will go into 1.7).

ryan-talley:
Sounds good I'll add to the release notes.

sjberman:
Scratch my last comment, just add it to the "next-release" section.

ryan-talley:
Got it.

docs/RELEASE-NOTES.rst
```diff
@@ -4,6 +4,11 @@ Release Notes for BIG-IP Controller for Kubernetes
 next-release
 ------------
 
+Bug Fixes
+`````````
+* :issues:`735` - Deleted rules from routes and ingresses on the same service not cleaned up properly
+* :issues:`753` - Issue 735 causing the controller to periodically delete and recreate ltm policy objects
```
sjberman: Let's move bug fixes below added functionality to keep the formatting consistent. Also, is 753 related to the url-rewrite issue? If so, let's not add that to the release notes in this PR, we can do that for the specific bug fix MR.

ryan-talley: Fixed

docs/RELEASE-NOTES.rst
```diff
@@ -9,6 +9,10 @@ Added Functionality
 * Added `--manage-configmaps` argument to CC to prevent or allow CC to respond to ConfigMap events. Defaults to `true`.
 * Added `virtual-server.f5.com/whitelist-source-range` annotation to support CIDR whitelisting.
 
+Bug Fixes
+`````````
+* :issues:`735` - Deleted rules from routes and ingresses on the same service not cleaned up properly
```
sjberman: Sorry, nitpick. Add a period to the end of the sentence.

ryan-talley:
Added

PR #761, opened 2018-10-04
WIP: Dependency management changes for annotation rules.

ryan-talley:


pkg/appmanager/appManager.go
```diff
@@ -1193,15 +1346,17 @@ func (appMgr *Manager) syncRoutes(
 			_, routeFound, _ := appInf.routeInformer.GetIndexer().GetByKey(routeKey)
 			return !routeFound
 		}
-		_, depsRemoved := appMgr.resources.UpdateDependencies(
+		depsAdded, depsRemoved := appMgr.resources.UpdateDependencies(
 			objKey, objDeps, svcDepKey, routeLookupFunc)
 
+		updateURL, updateApp := processAnnotationDependencies(depsAdded, depsAdded, false)
```
ryan-talley: depsRemved second arg

pkg/appmanager/resourceConfig.go
```diff
@@ -1452,7 +1483,7 @@ func (appMgr *Manager) createRSConfigFromIngress(
 				// the old value is cleaned up. All rules will be unmerged for the pool, but those
 				// that still remain will be re-merged before writing out the final config.
 				// Only deleted annotations will be completely removed.
-				appMgr.ProcessAnnotationRules(&cfg, newPool.Name)
+				//appMgr.ProcessAnnotationRules(&cfg, newPool.Name)
```
ryan-talley: Down below on 1497  where we check if rules already exist, but only if they are not annotation rules we go with the new one. This needs to be changed to now check for annotation rules and if they already exist use the old value, since that means the annotation  has not changed.

ryan-talley: We would also need to check the mergedRules and if that rule name is in it do not add the new rule

pkg/appmanager/appManager.go
```diff
@@ -1145,6 +1146,158 @@ func (appMgr *Manager) syncIngresses(
...
+	}
+
+	var urlUpdate bool
+	if urlAdded.Name != "" && urlRemoved.Name != "" {
+		urlUpdate = true
+		urlRemoved.Name = getRemovedAnnotationRuleNames(urlAdded.Name, urlRemoved.Name, false, obj)
+	} else if urlAdded.Name != "" && urlRemoved.Name == "" {
+		urlUpdate = true
+	} else if urlAdded.Name == "" && urlRemoved.Name != "" {
+		urlRemoved.Name = getRemovedAnnotationRuleNames("", urlRemoved.Name, false, obj)
```
sjberman: These urlRemoved and appRemoved aren't returned from the function; where are they actually used?

pkg/appmanager/resourceConfig.go
```diff
@@ -1512,10 +1543,6 @@ func (appMgr *Manager) createRSConfigFromIngress(
 	}
 
 	log.Debugf("%d whitelist sources", len(whitelistSourceRangeRefs))
-	if len(urlRewriteRefs) > 0 || len(appRootRefs) > 0 || len(whitelistSourceRangeRefs) > 0 {
-		cfg.MergeRules(appMgr.mergedRulesMap)
-	}
```
sjberman: Does this need to be removed? Will this cause us to miss the whitelistSourceRangeRefs case?

ryan-talley: I think this is fine, but I will talk to Tim Rupp about it, since merging rules only really matters when we have the url-rewrite or app-root annotations and this will be called at the end syncRoutes if those annotations are applied.

PR #764, opened 2018-10-09
Whitelist IP support for Routes

sjberman:
Adding support for the whitelist annotation for Routes. This is the same functionality as Ingresses (only allow specific IP access) for certain L7 rules.

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -1132,22 +1132,7 @@ func (appMgr *Manager) syncIngresses(
 					}
 				}
 				if dep.Kind == WhitelistDep {
-					for _, pol := range rsCfg.Policies {
-						for _, rl := range pol.Rules {
-							for i, cd := range rl.Conditions {
-								var valueStr string
-								for _, val := range cd.Values {
-									valueStr += val + ","
-								}
-								valueStr = strings.TrimSuffix(valueStr, ",")
-								if valueStr == dep.Name {
-									copy(rl.Conditions[i:], rl.Conditions[i+1:])
-									rl.Conditions[len(rl.Conditions)-1] = nil
-									rl.Conditions = rl.Conditions[:len(rl.Conditions)-1]
-								}
-							}
-						}
-					}
+					rsCfg.deleteWhitelistCondition(dep.Name)
 				}
 			}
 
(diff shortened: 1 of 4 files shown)


pkg/appmanager/appManager.go
```diff
@@ -1320,6 +1320,24 @@ func (appMgr *Manager) syncRoutes(
 						}
 					}
 				}
+				if dep.Kind == WhitelistDep {
```
ryan-talley: Since this is exactly the same as it is in syncIngress lets make this a common function between the two

pkg/appmanager/resourceConfig.go
```diff
@@ -2062,6 +2070,30 @@ func (appMgr *Manager) handleRouteTls(
...
 		rc.MergeRules(appMgr.mergedRulesMap)
 	}
+
+	// Add whitelist conditions to every rule (to prevent access)
+	var whitelistSourceRanges []string
+	if sourceRange, ok := route.ObjectMeta.Annotations[f5VsWhitelistSourceRangeAnnotation]; ok {
+		whitelistSourceRanges = parseWhitelistSourceRangeAnnotations(sourceRange)
+	}
+	if len(whitelistSourceRanges) > 0 {
+		for _, pol := range rc.Policies {
```
ryan-talley: Can we turn this block into a common function as well?

ryan-talley:
See my comments.

pkg/appmanager/resourceConfig.go
sjberman: This block is slightly unique to this function.

ryan-talley:
LGTM

pkg/appmanager/appManager_test.go
```diff
@@ -3484,6 +3484,80 @@ var _ = Describe("AppManager Tests", func() {
...
+						},
+					}
+					route := test.NewRoute("route", "1", namespace, spec,
+						map[string]string{
+							"virtual-server.f5.com/whitelist-source-range": "1.2.3.4/32,5.6.7.8/24",
+						})
+					mockMgr.addRoute(route)
+					spec2 := routeapi.RouteSpec{
+						Host: "bar.com",
+						Path: "/bar",
```
ryan-talley: Test

Issue #765, opened 2018-10-10
Route L7 rules are not sorted by URI


sjberman:
#### Description
Currently, the matching strategy for policies is hard coded to "first-match", because the controller is then supposed to sort the rules by URI length. This implements a more efficient "best-match" algorithm. Rules for Ingresses are sorted properly, however rules for Routes are not.

The current workaround is to create Routes in a specific order so that you manually sort the rules. The controller needs to sort these rules on its own.

#### Controller Version
1.6.1

PR #766, opened 2018-10-12
Add whitelist reset rule, fix Route sorting

### LINKED ISSUE #755: RFE: Ability to select the VServer Policy traffic matching strategy
#### Description
Currently, the VServer Policy traffic matching strategy is hardcoded to be "first-match". We would like the ability to change this to "best-match".

Use case is as follows: we have multiple applications being deployed onto OpenShift, with Route objects configured by the k8s-bigip-ctlr onto the F5. One of these apps is the "root" app, meaning it is available on /. Other apps are configured on /\<appname\>. 

In the first-match policy, we have to ensure that the root app is always the last rule in the policy. This can only be ensure by deleting and re-adding the route for the "root" app whenever any route is added, since route order is based on creationTimestamp of the Route object. This approach means that we need to remove and readd the main route object for our application on every deploy, causing a short outage while the k8s-bigip-ctlr removes and readds the rule.

Using the best-match policy, traffic is directed towards the correct pod regardless of the route order, and we no longer need to recreate the root app rule.

#### OpenShift Version
3.9

#### Controller Version
1.6.1

#### BIG-IP Version
13.1.10.3 Build 0.0.5 Point Release 3

#### Diagnostic Information
Currently, the first-match policy appears to be hardcoded [here](https://github.com/F5Networks/k8s-bigip-ctlr/blob/25739b34e1be6340e743a68fc1afb0e274ac96b2/pkg/appmanager/routing.go#L184)

sjberman:
Problem: 1. Route rules were not sorted correctly, causing "first-match" to behave unintentionally.
2. Whitelisted L7 rules for Routes with iRules would still allow traffic to flow through the iRule even when failing the policy.
3. Policy with whitelist rules didn't have the required "tcp" flag.

Solution: 1. Sorted Routes by URI to mimic "best-match".
2. Whitelisted Route rules now add a "reset connection" rule to prevent traffic from unintentionally flowing through an iRule.
3. Added "tcp" field to policy.

Fixes #755 #765

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -14,6 +14,8 @@ Bug Fixes
 `````````
 * :issues:`735` - Deleted rules from routes and ingresses on the same service not cleaned up properly.
 * :issues:`753` - Controller doesn't delete and recreate annotation-based policy rules.
+* :issues:`755` - Controller implements best-match by setting first-match and sorting rules in reverse lexical order.
+* :issues:`765` - Controller properly sorts Route rules in reverse lexical order.
 
 v1.6.1
 ------
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -1266,7 +1267,7 @@ func (appMgr *Manager) syncRoutes(
 				}
 			}
 
-			// Remove any left over pools from services no longer used by this Route
+			// Remove any left over pools/rules from configs no longer used by this Route
 			for _, dep := range depsRemoved {
 				if dep.Kind == ServiceDep {
 					cfgChanged, svcKey := rsCfg.RemovePool(
@@ -1280,15 +1281,19 @@ func (appMgr *Manager) syncRoutes(
 				}
 				if dep.Kind == RuleDep {
 					for _, pol := range rsCfg.Policies {
+						var toRemove []*Rule
 						for _, rl := range pol.Rules {
 							if rl.FullURI == dep.Name {
-								rsCfg.DeleteRuleFromPolicy(pol.Name, rl, appMgr.mergedRulesMap)
+								toRemove = append(toRemove, rl)
 								// Delete profile (route only)
 								if rsCfg.MetaData.ResourceType == "route" {
 									resourceName := strings.Split(rl.Name, "_")[3]
 									rsCfg.DeleteRouteProfile(dep.Namespace, resourceName)
 								}
 							}
+						}
+						for _, rl := range toRemove {
+							rsCfg.DeleteRuleFromPolicy(pol.Name, rl, appMgr.mergedRulesMap)
 						}
 					}
 				}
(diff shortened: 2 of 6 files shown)


docs/RELEASE-NOTES.rst
```diff
@@ -14,6 +14,8 @@ Bug Fixes
 `````````
 * :issues:`735` - Deleted rules from routes and ingresses on the same service not cleaned up properly.
 * :issues:`753` - Controller doesn't delete and recreate annotation-based policy rules.
+* :issues:`755` - Controller implements best-match by setting first-match and sorting rules by URI length.
```
f5yacobucci: Outside of the madness that's used for comparing the approot and urlrewrite rules, we're technically sorting in reverse lexical order and not simply URI length.

pkg/appmanager/appManager.go
```diff
@@ -1280,16 +1281,20 @@ func (appMgr *Manager) syncRoutes(
...
+								toRemove = append(toRemove, rl)
 								// Delete profile (route only)
 								if rsCfg.MetaData.ResourceType == "route" {
 									resourceName := strings.Split(rl.Name, "_")[3]
 									rsCfg.DeleteRouteProfile(dep.Namespace, resourceName)
 								}
 							}
 						}
+						for _, rl := range toRemove {
+							rsCfg.DeleteRuleFromPolicy(pol.Name, rl, appMgr.mergedRulesMap)
```
f5yacobucci: Does this have to be removed after looping over the policy rules? It can't be removed inline?

pkg/appmanager/appManager.go
```diff
@@ -1309,6 +1314,18 @@ func (appMgr *Manager) syncRoutes(
 					rsCfg.deleteWhitelistCondition(dep.Name)
 				}
 			}
+			// Sort the rules and reset their ordinals
+			sortrules := func(rls *Rules, ordinal int) {
+				sort.Sort(sort.Reverse(*rls))
+				for _, v := range *rls {
```
f5yacobucci: You can save this loop by updating the ordinal during the sort: https://play.golang.org/p/_DENN5qxR7e

pkg/appmanager/resourceConfig.go
```diff
@@ -2071,25 +2071,57 @@ func (appMgr *Manager) handleRouteTls(
...
+
+						// Add reset traffic rule immediately after this rule
+						if (len(pol.Rules) > i+1 && pol.Rules[i+1].Name != rl.Name+"-reset") ||
+							i == len(pol.Rules)-1 {
+							reset := &Rule{
+								Name:    rl.Name + "-reset",
+								FullURI: rl.FullURI,
+								Actions: []*action{{
+									Name:    "0",
+									Forward: true,
```
f5yacobucci: Is forward really required on a reset action?

pkg/appmanager/resourceConfig.go
```diff
@@ -2071,25 +2071,57 @@ func (appMgr *Manager) handleRouteTls(
...
 			if pol.Name == policyName {
-				for _, rl := range pol.Rules {
-					cond := condition{
-						Tcp:     true,
-						Address: true,
-						Matches: true,
-						Name:    "0",
-						Values:  whitelistSourceRanges,
+				for i, rl := range pol.Rules {
+					if rl.FullURI == rule.FullURI && !strings.HasSuffix(rl.Name, "-reset") {
```
f5yacobucci: Is this run before or after sorting?

If before, do we always get the correct ordering, I have a test that will sometimes reorder equal strings.

If after, should we be updating ordinals on the previously set rules?

pkg/appmanager/resourceConfig.go
sjberman: Yep.

pkg/appmanager/resourceConfig.go
sjberman: This is run before sorting.

f5yacobucci: I'm worried we might move these out of order, if you run this: https://play.golang.org/p/7bRNfjF4SRF you'll see the original ordering doesn't always get preserved. Moving the original position of the quux.com strings effects the final position.

PR #767, opened 2018-10-13
Changed to debian base container instead of alpine

amudukutore:


sjberman:
There are a couple of additional things that can be removed in relation to the debug container.

https://github.com/F5Networks/k8s-bigip-ctlr/blob/master/build-tools/_build-lib.sh#L23
Makefile debug/dbg-build step
build-tools/build-debug-artifacts.sh
build-tools/dbg-build.sh

PR #811, opened 2019-01-25
Delay vsQueue Length Assessment

### LINKED ISSUE #810: Controller deletes services and recreates during bigip-ctlr pod restart
#### Description
 Whenever the CC Pod crashes (or during scheduled upgrades) - the entire partition it manages gets wiped out - then it a couple of minutes to restore all the services configured in K8s to get configured on the BIG-IP again. This is occurring when a large number of services configured in k8s.

#### Kubernetes Version
1.10

#### Controller Version
1.7.0

#### BIG-IP Version
13.1

ssurenr:
Mitigates #810 where the controller deletes BigIP resources on startup and restart where a large number of services present in Kubernetes

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -765,6 +765,8 @@ func (appMgr *Manager) virtualServerWorker() {
 func (appMgr *Manager) processNextVirtualServer() bool {
 	key, quit := appMgr.vsQueue.Get()
 	if !appMgr.initialState && appMgr.processedItems == 0 {
+		//TODO: Properly handlle queueLen assessment and remove Sleep function
+		time.Sleep(1 * time.Second)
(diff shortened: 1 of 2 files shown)


pkg/appmanager/appManager.go
```diff
@@ -765,6 +765,8 @@ func (appMgr *Manager) virtualServerWorker() {
 func (appMgr *Manager) processNextVirtualServer() bool {
 	key, quit := appMgr.vsQueue.Get()
 	if !appMgr.initialState && appMgr.processedItems == 0 {
+		//TODO: Properly handlle queueLen assessment and remove Sleep function
+		time.Sleep(1 * time.Second)
```
sjberman: This seems very arbitrary. What are the results now when testing with this sleep?

sjberman:
Ensure the pipeline passes, and then you're good to merge.

Issue #809, opened 2019-01-23
building error pip no such option: --process-dependency-links


vincentmli:
#### Description
<For bug: Describe the bug in detail, observed versus expected behavior, steps to reproduce the issue, and a brief description of your deployment scenario>

I had development environment setup for the controller in Macbook pro, I accidentally removed all the docker images including the python-slim container image, when I run 'make prod' again, I experienced a few errors as below:
<pre>
Installing collected packages: configparser, enum34, pyflakes, mccabe, pycodestyle, flake8, virtualenv
  Running setup.py install for configparser: started
    Running setup.py install for configparser: finished with status 'done'
Successfully installed configparser-3.5.0 enum34-1.1.6 flake8-3.6.0 mccabe-0.6.1 pycodestyle-2.4.0 pyflakes-2.0.0 virtualenv-16.2.0

Usage:   
  pip install [options] <requirement specifier> [package-index-options] ...
  pip install [options] -r <requirements file> [package-index-options] ...
  pip install [options] [-e] <vcs project url> ...
  pip install [options] [-e] <local project path> ...
  pip install [options] <archive url/path> ...

**no such option: --process-dependency-links**
Removing intermediate container c8b688dd4ac7
The command '/bin/sh -c pip install --no-cache-dir --upgrade pip &&     pip install --no-cache-dir setuptools flake8 virtualenv && 	pip install --no-cache-dir --process-dependency-links -r /tmp/requirements.txt && 	pip install --no-cache-dir -r /tmp/requirements.docs.txt && 	go get github.com/wadey/gocovmerge && 	go get golang.org/x/tools/cmd/cover && 	go get github.com/mattn/goveralls && 	go get github.com/onsi/ginkgo/ginkgo && 	go get github.com/onsi/gomega && 	chmod 755 /entrypoint.sh' returned a non-zero code: 2
make: *** [prod-build] Error 2

Successfully installed configparser-3.5.0 enum34-1.1.6 flake8-3.6.0 mccabe-0.6.1 pycodestyle-2.4.0 pyflakes-2.0.0 virtualenv-16.2.0
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
[...]

PR #813, opened 2019-01-25
Fix process-dependency-links removal from pip

ssurenr:
Fixes #812


vincentmli:
Hi Surendhar,

have you tried to build the CC image with this changes and run the CC in cluster mode and deploy an ingress? I want to confirm if you could get error I had in  https://github.com/F5Networks/k8s-bigip-ctlr/issues/809#issuecomment-461278382, something appears to be broken with above changes, it complains cccl-net-api-schema.yml missing

ssurenr:
Changes merged in other PR

Issue #802, opened 2018-12-19
Rule http_redirect_irule_443 or datagroup https_redirect_dg do not respect virtual-server.f5.com/http-port


bogiva:
#### Description

Neither the code at https://github.com/F5Networks/k8s-bigip-ctlr/blob/ae907570e02b6149ba6a636685fa1b9c53ea2eb9/pkg/appmanager/routing.go#L353 nor the code at https://github.com/F5Networks/k8s-bigip-ctlr/blob/55007560d27df820f1d5cea7ffd5b69599912d3a/pkg/appmanager/appManager.go#L977
do respect the value of the "virtual-server.f5.com/http-port" property.

If we set this property e.g. to 8080 then the redirect irule mentioned above does not work as expected, i.e. does not redirect to HTTPS.

#### Kubernetes Version
1.10.10

#### Controller Version
1.4.2

#### BIG-IP Version
12.1.2

vincentmli:
according to https://devcentral.f5.com/wiki/iRules.getfield.ashx

[HTTP::host]  could be  <hostname>:<port>


maybe change 

<pre>                               
 if {$redir == 1} {
                                        HTTP::redirect https://[getfield [HTTP::host] ":" 1]:443[HTTP::uri]
                                }

</pre>
to

<pre>
                                if {$redir == 1} {
                                        HTTP::redirect https://[HTTP::host]:443[HTTP::uri]
                                }
</pre>

vincentmli:
I did a test with following ingress using 'virtual-server.f5.com/http-port': '8080' and the redirect works fine, could you please let me know what I am missing here? any further detail from you?

<pre>
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: single-service
  annotations:
     ingress.kubernetes.io/ssl-redirect: "true"
     ingress.kubernetes.io/allow-http: "false"
     ingress.kubernetes.io/allow-https: "true"
     kubernetes.io/ingress.class: f5
     'virtual-server.f5.com/ip': '10.169.72.169'
     'virtual-server.f5.com/http-port': '8080'
spec:
  tls:
#  - secretName: test/test-clientssl
  - secretName: Common/clientssl
  backend:
    serviceName: nginxservice 
    servicePort: 80
</pre>

I have no problem with http_redirect_irule_443, my test with curl to visit 10.169.72.169:8080 and it got  redirect to https 443 virtual server as the alert from the certificate error

<pre>
[...]

PR #816, opened 2019-02-07
http_redirect_irule_443 irule should honor custom http port

vincentmli:
there are actually two issues here when Ingress ssl-redirect is true

1, the http virtual server should not have policy configuation as the
policy would handle the request instead of ssl-redirect
2, when http-port is non standard port 80, the ssl-redirect irule should
take non-standard port 80 into consideration

this patch fixes above two issues


vincentmli:
this PR address https://github.com/F5Networks/k8s-bigip-ctlr/issues/802

PR #821, opened 2019-02-13
Fix Builds due to process-dependency-links removal

ssurenr:
Updated conf.py to make compatible with Sphinx version 1.8.3
Added f5-cccl external dependency to requirements.txt
Change as per go fmt
Change Docker Repo for attrib generator

### DIFF
--- a/build-tools/attributions-generator.sh
+++ b/build-tools/attributions-generator.sh
@@ -7,7 +7,7 @@
 set -e
 set -x
 
-ATTR_GEN_IMG=f5devcentral/attributions-generator:latest
+ATTR_GEN_IMG=f5networksdevel/attributions-generator:latest
 docker pull ${ATTR_GEN_IMG}
 
 RUN_ARGS=( \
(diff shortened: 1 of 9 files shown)


build-tools/attributions-generator.sh
```diff
@@ -7,7 +7,7 @@
 set -e
 set -x
 
-ATTR_GEN_IMG=f5devcentral/attributions-generator:latest
+ATTR_GEN_IMG=f5networksdevel/attributions-generator:latest
```
amudukutore: @ssurenr , @veeramalla-f5 - The f5networksdevel repo should not be used for any public dependencies. You should upload the latest attributions-generator image from there into f5devcentral.

requirements.txt
```diff
@@ -1 +1,2 @@
 -e git+https://github.com/f5devcentral/f5-ctlr-agent.git@671a0defbea9e7def753adb685cbd9230754f131#egg=f5-ctlr-agent
+-e git+https://github.com/f5devcentral/f5-cccl.git@cfaaa7713fddf9fe28400317548d25b104d3b865#egg=f5-cccl
```
amudukutore: At one point, the CC team intentionally removed direct dependency between the controller and cccl, and instead added a direct dependency only on f5-ctlr-agent. I expect you have to do this since the process-dependency-links option is no longer supported. Perhaps, it is worth to see how multi-level dependencies can be leveraged without calling them out at the top level.

ssurenr: We will create a backlog item to evaluate a permanent solution for this issue.

build-tools/attributions-generator.sh
ssurenr: We kept it here due to the access restrictions. We will move this back to f5devcentral shortly.

Issue #797, opened 2018-12-15
vxlan manager only use node internal IP for fdb endpoint IP cause wrong FDB in bigip


myf5:
#### Description
Here is my k8s cluster environment:

1. Set k8s cluster uses ens33 (172.16.10.0/24) for k8s cluster everything(certificates, Node internalIP, /etc/hosts setting)
2. Set flannel to use ens160, for vxlan, so flannel public-ip is using 172.16.40.0/24 net. This is different to the node internal IP net.

Nodes interfaces:

```
//k8s master//
ens33: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 172.16.10.201  netmask 255.255.0.0  broadcast 172.16.255.255
        inet6 fe80::93eb:77c9:718:43d2  prefixlen 64  scopeid 0x20<link>
        ether 00:50:56:b3:46:85  txqueuelen 1000  (Ethernet)
        RX packets 352566  bytes 32062651 (30.5 MiB)
        RX errors 0  dropped 4  overruns 0  frame 0
        TX packets 457619  bytes 416423360 (397.1 MiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

ens160: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 172.16.40.199  netmask 255.255.255.0  broadcast 172.16.40.255
        inet6 fe80::a35:8f62:68df:ae99  prefixlen 64  scopeid 0x20<link>
        ether 00:50:56:b3:09:f2  txqueuelen 1000  (Ethernet)
        RX packets 278  bytes 26665 (26.0 KiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 159  bytes 19746 (19.2 KiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

flannel.1: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1450
        inet 10.244.0.0  netmask 255.255.255.255  broadcast 0.0.0.0
        inet6 fe80::7046:9bff:fe13:8814  prefixlen 64  scopeid 0x20<link>
        ether 72:46:9b:13:88:14  txqueuelen 0  (Ethernet)
        RX packets 80  bytes 7731 (7.5 KiB)
        RX errors 0  dropped 0  overruns 0  frame 0
[...]

PR #827, opened 2019-02-14
Use flannel.alpha.coreos.com/public-ip as node endpoint

vincentmli:
when node has multiple node ip and  CC deployed with flannel, CC
could select wrong node ip as net fdb endpoint in BIGIP, which result
in no network connectivity between BIGIP and POD. Since CC already
use flannel.alpha.coreos.com/backend-data for VtepMAC, for accuracy,
CC should also use flannel.alpha.coreos.com/public-ip as net fdb endpoint
for the node. this resolves issue https://github.com/F5Networks/k8s-bigip-ctlr/issues/797


ssurenr:
@veeramalla-f5 Please have a look and close if already implemented.

veeramalla-f5:
Closing this pull request as the fix is addressed as part of pull request #797 and merged into controller in the release 1.9.1.

PR #844, opened 2019-03-05
Fix security vulnerabilities

nandakishorepeddi:
Fix security vulnerabilities

    Problem:
    Python dependencies are having vulnerabilities which affect https
    communication between controller and dependencies

    Solution:
    Upgrade python libraries to the patched versions and tested the
    functionalities.

    affects-branches: master, 1.8-stable


ssurenr:
Addressed in #845

PR #849, opened 2019-03-06
Fix process-dependency-links removal from pip

papineni87:
Updated conf.py to make compatible with Sphinx version 1.8.3
Added f5-cccl external dependency to requirements.txt
Change as per go fmt
Change Docker Repo for attrib generator


ssurenr:
Could you add approvals as well? This will pass the build.

papineni87:
Required changes are there in 1.8-stable, so  closing the pull request

PR #854, opened 2019-03-12
Branch out the CC implementation for AS3 Specific implementation.

trinaths:
This fix processes configmaps from the work-queue.

1. appManager and Validate Resources are updated with logic to branchout.

### DIFF
--- a/pkg/appmanager/validateResources.go
+++ b/pkg/appmanager/validateResources.go
@@ -29,12 +29,25 @@ func (appMgr *Manager) checkValidConfigMap(
 ) (bool, []*serviceQueueKey) {
 	// Identify the specific service being referenced, and return it if it's
 	// one we care about.
+	var keyList []*serviceQueueKey
 	cm := obj.(*v1.ConfigMap)
 	namespace := cm.ObjectMeta.Namespace
 	_, ok := appMgr.getNamespaceInformer(namespace)
 	if !ok {
 		// Not watching this namespace
 		return false, nil
+	}
+	//check as3 config map
+	if ok := appMgr.checkAs3ConfigMap(obj); ok {
+		log.Debugf("[as3_log] Found AS3: %t. Processing ConfigMap - %s with Template %s",
+			ok, cm.ObjectMeta.Name, cm.Data)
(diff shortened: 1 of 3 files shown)


pkg/appmanager/validateResources.go
```diff
@@ -29,13 +29,26 @@ func (appMgr *Manager) checkValidConfigMap(
...
+	var keyList []*serviceQueueKey
 	cm := obj.(*v1.ConfigMap)
 	namespace := cm.ObjectMeta.Namespace
 	_, ok := appMgr.getNamespaceInformer(namespace)
 	if !ok {
 		// Not watching this namespace
 		return false, nil
 	}
+	//check as3 config map
+	if as3ok := appMgr.checkAs3ConfigMap(obj); as3ok {
```
ssurenr:
```suggestion
	if ok = appMgr.checkAs3ConfigMap(obj); ok {
```
We can reuse the same `ok` variable.

pkg/appmanager/validateResources.go
```diff
@@ -363,3 +375,30 @@ func validateAppRootAnnotations(rsType int, entries map[string]string) {
...
+) bool {
+	// check for metadata.labels has 'as3' and that 'as3' is set to 'true'
+	cm := obj.(*v1.ConfigMap)
+	labels := cm.ObjectMeta.Labels
+	log.Debugf("[as3_log] Parsing labels: %s ", labels)
+	if val, ok := labels["as3"]; ok {
+		log.Debugf("[as3_log] Found AS3 config map...")
+		if val == "true" {
+			return true
+		} else {
```
ssurenr: https://github.com/F5Networks/k8s-bigip-ctlr/blob/71ffc366140fab51af0951ab6017a738739dfd8c/pkg/appmanager/validateResources.go#L395-L403
Let's replace this block with a single `return false`.

ssurenr: log at line 400 is of less significance

pkg/appmanager/validateResources.go
trinaths: ok. Taken care

PR #855, opened 2019-03-13
Add AS3 Manager Design

ssurenr:
Add skeleton for developing user stories.


ssurenr:
> Please add
> type service []serviceName
> 
> This is used to while fetching list of services

You mean `type application []serviceName`?

PR #867, opened 2019-03-19
POST as3 declaration to BigIP

agupta-F5:
This code will POST AS3 declaration to Big-IP.

    1. Implemented function postAS3Declaration to POST AS3 declaration to Big IP.
    2. Created new generic function restCallToBigIP to handle all REST calls.
    3. Created new file as3Manager_test.go to write unit test cases.

### DIFF
--- a/pkg/appmanager/as3Manager.go
+++ b/pkg/appmanager/as3Manager.go
@@ -305,11 +305,11 @@ func (as3RestClient *As3RestClient) restCallToBigIP(method string, route string,
 	}
 	req.SetBasicAuth(BigIPUsername, BigIPPassword)
 	resp, err := as3RestClient.client.Do(req)
-	defer resp.Body.Close()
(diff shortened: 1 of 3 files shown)


pkg/appmanager/as3Manager.go
```diff
@@ -252,7 +269,52 @@ func (appMgr *Manager) buildAS3Declaration(obj as3Object, template as3Template)
...
+	if err != nil {
+		log.Errorf("[as3_log] REST call error: %v ", err)
+		return string(body), false
+	}
+	body, err = ioutil.ReadAll(resp.Body)
+	if err != nil {
+		log.Errorf("[as3_log] REST call error: %v ", err)
+		return string(body), false
+	}
+	defer resp.Body.Close()
```
subbuv26: defer should be moved to where the "resp" gets created

PR #869, opened 2019-03-22
Add Unit tests for AS3Manager

subbuv26:
1. Added Unit tests for AS3Manager
2. Small tweak to handle "generic" AS3 services

### DIFF
--- a/pkg/appmanager/as3Manager.go
+++ b/pkg/appmanager/as3Manager.go
@@ -12,8 +12,7 @@ import (
 
 	log "github.com/F5Networks/k8s-bigip-ctlr/pkg/vlogger"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	apiv1 "k8s.io/client-go/pkg/api/v1"
-	v1 "k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/pkg/api/v1"
 )
 
 type as3Template string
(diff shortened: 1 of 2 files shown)


pkg/appmanager/as3Manager.go
```diff
@@ -91,8 +91,8 @@ func (appMgr *Manager) getAS3ObjectFromTemplate(
 					continue
 				}
 
-				// filter out empty json objects and pool objects
-				if cl := getClass(v); cl == "" || cl == "Pool" {
+				// filter out pool objects
+				if cl := getClass(v); cl == "Pool" {
```
ssurenr: Could you make this a separate commit with a proper commit mentioning what problem it addresses? It will be easy for us to look back in the future.

pkg/appmanager/as3Manager_test.go
```diff
@@ -19,30 +19,83 @@ import (
...
+				RouteClientV1:    test.CreateFakeHTTPClient(),
+				IsNodePort:       true,
+				broadcasterFunc:  NewFakeEventBroadcaster,
+				ManageConfigMaps: true,
+			})
+		})
+		AfterEach(func() {
+			mockMgr.shutdown()
+		})
+
```
ssurenr: Could you add a test case that passes a JSON file with errors and expects this error? This should make the coverage 100%
https://github.com/F5Networks/k8s-bigip-ctlr/blob/794b182fe2756db550fda700fd76579b2679a177/pkg/appmanager/as3Manager.go#L63

PR #871, opened 2019-03-28
This commit validates as3 template using as3 schema

veeramalla-f5:
This commit validates  as3 template using as3 schema

Known Issues: NA

Overall Changes:
1. Updated the gojsonschema to latest version.
2. Made below changes to AS3 schema.
           1. Changed the minLimit of serverAddress from 1 to 0.
           2. Changed the maximum repeat count from 65520 and 8192 to 1000.
           3. Changed the URN reference of "$id" to URL.
3. Updated the build files to push as3 schema to container along with existing schemas. 
4. Introduced a new function to as3Manager.go(validateAS3) to validate schema.

Unit Tests: Passed
Regression: Passed

Controller Image created with these changes -> abhishekf5/k8s-bigip-ctlr:as3schemavalidation-1000

### DIFF
--- a/pkg/appmanager/as3Manager.go
+++ b/pkg/appmanager/as3Manager.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	log "github.com/F5Networks/k8s-bigip-ctlr/pkg/vlogger"
+	"github.com/xeipuuv/gojsonschema"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/client-go/pkg/api/v1"
 )
@@ -54,6 +55,15 @@ var BigIPURL string
 
 // Takes an AS3 Template and perform service discovery with Kubernetes to generate AS3 Declaration
 func (appMgr *Manager) processUserDefinedAS3(template string) bool {
+
+	// Validate AS3 Template
+	ok := appMgr.validateAS3Template(template)
+
+	if !ok {
+		log.Errorf("Error processing AS3 template \n")
+		return false
+	}
+
 	templateObj := as3Template(template)
 	obj, ok := appMgr.getAS3ObjectFromTemplate(templateObj)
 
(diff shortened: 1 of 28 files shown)


pkg/appmanager/as3Manager.go
```diff
@@ -52,6 +63,42 @@ func (appMgr *Manager) processUserDefinedAS3(template string) bool {
 	return true
 }
 
+// Validates the AS3 Template
+func (appMgr *Manager) validateAS3Template(template string) bool {
+
+	workingDir, _ := os.Getwd()
+	var schemaUrl = workingDir + "/vendor/src/f5/schemas/as3-schema-3.9.0-3-cis.json"
```
ssurenr: Suggested use of `appMgr.schemaLocal`

https://github.com/F5Networks/k8s-bigip-ctlr/blob/f1e115ed7a1ccce2767df9b6afc9fe28d9ffb80b/pkg/appmanager/resourceConfig.go#L1285

pkg/appmanager/as3Manager.go
```diff
@@ -52,6 +63,42 @@ func (appMgr *Manager) processUserDefinedAS3(template string) bool {
...
+
+	var schema, err = ioutil.ReadFile(schemaUrl)
+
+	if err != nil {
+		log.Errorf("Schema not found :\n")
```
ssurenr:
```suggestion
		log.Errorf("Error opening schema file :\n %v", err)
```

pkg/appmanager/as3Manager.go
```diff
@@ -52,6 +63,42 @@ func (appMgr *Manager) processUserDefinedAS3(template string) bool {
...
+
+	if result.Valid() {
+		log.Debugf("AS3 Template is Validated Successfully \n")
+		return true
+	} else {
+		log.Errorf("AS3 Template is not valid. see errors :\n")
+		for _, desc := range result.Errors() {
+			log.Errorf("- %s\n", desc)
+		}
+		return false
```
ssurenr: Suggestion: It is a good practice to look for failure block with if condition and put the true block out of `if` check.

```golang
if !result.Valid() {
  log.Errorf("Error")
  return false
}

return true
```

PR #873, opened 2019-03-28
Handle AS3 Templates with no ADC class.

trinaths:
Problem:
Without ADC class also, AS3 Template is being parsed and Controller is crashing.

Solution:
Updated logic not to handle any AS3 Template without ADC class.

All related unit tests pass.

affects-branches: feature.user-defined-as3


ssurenr:
Please resolve the conflicts.

trinaths:
Moving with new commit with resolved conflicts.

PR #881, opened 2019-04-05
Option to disable AS3 Validation

ssurenr:
Problem: We are customizing and hard linking AS3 JSON Schema with the
controller. If there are bugs in the JSON schema or the validator we use,
there is no way to address this situation without shipping a new release.

Solution: Make AS3 Schema validation configurable so that the user can
disable validation.

affected-branches: feature.user-defined-as3

### DIFF
--- a/cmd/k8s-bigip-ctlr/main.go
+++ b/cmd/k8s-bigip-ctlr/main.go
@@ -105,6 +105,7 @@ var (
 	bigIPPassword   *string
 	bigIPPartitions *[]string
 	credsDir        *string
+	as3Validation   *bool
 
 	vxlanMode        string
 	openshiftSDNName *string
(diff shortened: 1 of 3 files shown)


cmd/k8s-bigip-ctlr/main.go
```diff
@@ -172,6 +173,8 @@ func _init() {
 	credsDir = bigIPFlags.String("credentials-directory", "",
 		"Optional, directory that contains the BIG-IP username, password, and/or "+
 			"url files. To be used instead of username, password, and/or url arguments.")
+	as3Validation = bigIPFlags.Bool("as3-validation", true,
+		"Optional, specify whether or not to perform AS3 template validation before sending to BIG-IP.")
```
veeramalla-f5: "Optional, When set to false, disables as3 schema validation"

PR #882, opened 2019-04-08
Option to enable insecure communication

agupta-F5:
Problem: Insecure flag was always set to true.

Solution: Introducing --insecure flag option. The default option is set as false for secure communication. User can change it as true to enable insecure communication to bigip.

affects-branches: feature.user-defined-as3

### DIFF
--- a/pkg/appmanager/as3Manager.go
+++ b/pkg/appmanager/as3Manager.go
@@ -384,20 +384,20 @@ func (appMgr *Manager) postAS3Declaration(declaration as3Declaration) {
 	log.Debugf("[as3_log] Processing AS3 POST call with AS3 Manager")
 	var as3RC As3RestClient
 	as3RC.baseURL = BigIPURL
-	response, _ := as3RC.restCallToBigIP("POST", "/mgmt/shared/appsvcs/declare", declaration)
+	response, _ := as3RC.restCallToBigIP("POST", "/mgmt/shared/appsvcs/declare", declaration, appMgr.sslInsecure)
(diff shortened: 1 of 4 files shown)


pkg/appmanager/as3Manager.go
```diff
@@ -381,23 +381,23 @@ func (appMgr *Manager) buildAS3Declaration(obj as3Object, template as3Template)
 
 // Takes AS3 Declaration and post it to BigIP
 func (appMgr *Manager) postAS3Declaration(declaration as3Declaration) {
-	log.Debugf("[as3_log] Processing AS3 POST call with AS3 Manager")
+	log.Debugf("[as3_log] Processing AS3 POST call with AS3 Manager with insecure flag: %v", appMgr.sslInsecure)
```
ssurenr:
```suggestion
	log.Debugf("[as3_log] Processing AS3 POST call with AS3 Manager")
```

pkg/appmanager/as3Manager_test.go
```diff
@@ -103,7 +103,7 @@ var _ = Describe("As3Manager Tests", func() {
 			defer server.Close()
 			// Use Client & URL from our local test server
 			api := As3RestClient{server.Client(), server.URL}
-			_, status := api.restCallToBigIP(method, route, template)
+			_, status := api.restCallToBigIP(method, route, template, true)
```
ssurenr:
```suggestion
			_, status := api.restCallToBigIP(method, route, template, false)
```
Let's keep this to false as it is the ideal state.

PR #885, opened 2019-04-08
Update Limitations section

kkinch:
Update the Limitations for AS3. Hopefully this will be slightly more specific regarding:
 - Pool declaration (and the members in the pool).
 - ConfigMap instances (running state vs type of file).

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -6,14 +6,14 @@ next-release
 
 Added Functionality
 ```````````````````
-* Added support for `Application Services 3 <https://clouddocs.f5.com/products/extensions/f5-appsvcs-extension/latest>_` declarations.
+* Added support for `Application Services 3 Extension <https://clouddocs.f5.com/products/extensions/f5-appsvcs-extension/latest/>_`.
 
 Limitations
 ```````````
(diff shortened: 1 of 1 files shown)


docs/RELEASE-NOTES.rst
```diff
@@ -6,14 +6,14 @@ next-release
...
 Limitations
 ```````````
-* AS3 declaration with multiple members inside pool class are not supoorted.
-* Supports only one AS3 ConfigMap.
-* AS3 doesn't allow to move a node in BIGIP from old partition to a new partition.
-* Deleting AS3 ConfigMap doesnot delete static ARP entries.
+* AS3 pool class declarations support only one load balancing pool.
+* The BIG-IP Contoller supports only one AS3 ConfigMap instance.
+* AS3 does not support moving BIG-IP nodes to new partitions.
+* Static ARP entries remain after AS3 ConfigMaps are deleted.
```
ssurenr: Can we change this to active voice? CI is complaining.
```
In docs/RELEASE-NOTES.rst
=============
ain after AS3 ConfigMaps are deleted.
                         ^^^^^^^^^^^
"are deleted" may be passive voice on line 16 at column 49
```

PR #905, opened 2019-04-19
Change official build steps to reference Debian

pdp2shirts:
Changed text in Official Build steps to reference Debian rather than Alpine as we no longer build on an Alpine base image


pdp2shirts:
Not sure if it makes sense to cherry pick back to 1.9, 1.8, 1.7 since this change was made in 1.7

PR #907, opened 2019-04-30
Add event informers for ConfigMap create/update/delete events

subbuv26:


pkg/appmanager/as3Manager.go
```diff
@@ -435,3 +441,46 @@ func (as3RestClient *As3RestClient) restCallToBigIP(method string, route string,
...
 }
+
+// SetupAS3Informers returns an appInformer that includes the following set of informers.
+// CfgMapInformer and SvcInformer are label based and endptInformer is not label based.
+// These informers are event based informer and do not poll on the resources.
+func (appMgr *Manager) SetupAS3Informers() error {
+	// resyncPeriod is zero to avoid repolling
+	var resyncPeriod time.Duration
+	// namespace is Empty to create watchers for all namespaces
+	namespace := ""
```
ssurenr: There is variable `v1.NamespaceAll` which we can use.

PR #909, opened 2019-05-06
Fix797: Take the flannel node IP as Vtep endpoint

veeramalla-f5:
Problem: CIS takes the node ip information from kube api in cluster mode. It should actually take these details from flannel as CIS creates the fdb entries in BIG-IP using these details.

Solution: Use Flannel annotations to get the node ip addresses.

Affected Branches: master


ssurenr:
Resolves #797

PR #910, opened 2019-05-07
Implementation of sending AS3 declaration to bigip only if there is change in declaration

agupta-F5:
Problem:
CIS always posting same AS3 declaration to BIGIP if there are no changes in the declaration.

Solution:
Implemented MD5 checksum mechanism on the AS3 declaration to identify the changes on the declaration.

Affected Branches:
master

### DIFF
--- a/pkg/appmanager/as3Manager.go
+++ b/pkg/appmanager/as3Manager.go
@@ -18,6 +18,7 @@ package appmanager
 
 import (
 	"bytes"
+	"crypto/md5"
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


pkg/appmanager/as3Manager.go
```diff
@@ -369,18 +373,24 @@ func (appMgr *Manager) buildAS3Declaration(obj as3Object, template as3Template)
...
+	hash := md5.New()
+	io.WriteString(hash, string(declaration))
+	as3RestClient.newChecksum = string(hash.Sum(nil))
+	timeout := time.Duration(60 * time.Second)
 	var body []byte
+	if as3RestClient.oldChecksum == as3RestClient.newChecksum {
+		log.Debugf("[as3_log] No change in declaration.")
+		return string(body), true
+	}
+	as3RestClient.oldChecksum = as3RestClient.newChecksum
```
ssurenr: This needs to be done after we send the declaration to BIG-IP and it is successful.

PR #918, opened 2019-05-17
Option to add BIG-IP third party certificate/self signd certificate (using config map) to CIS

agupta-F5:
Problem:
CIS was not supporting third party certificates/self-signed certificates for secure communication to BIG-IP
Solution:
Given an option to pass the BIG IP's third-party certificates/self-signed certificates using config map for secure communication

Affected Branches:
master

### DIFF
--- a/pkg/appmanager/as3Manager.go
+++ b/pkg/appmanager/as3Manager.go
@@ -20,11 +20,13 @@ import (
 	"bytes"
 	"crypto/md5"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
 
 	log "github.com/F5Networks/k8s-bigip-ctlr/pkg/vlogger"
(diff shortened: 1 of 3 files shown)


pkg/appmanager/as3Manager.go
```diff
@@ -452,6 +471,22 @@ func (as3RestClient *As3RestClient) restCallToBigIP(method string, route string,
 
 }
 
+// Read certificate from configmap
+func (appMgr *Manager) getCertFromConfigMap(cfgmap string) {
+
+	certs := ""
+	cm, err := appMgr.kubeClient.CoreV1().ConfigMaps("default").Get(cfgmap, metaV1.GetOptions{})
```
ssurenr: This will only look for config maps in "default" namespace. Please change it to `v1.NamespaceAll`

pkg/appmanager/as3Manager.go
```diff
@@ -452,6 +471,22 @@ func (as3RestClient *As3RestClient) restCallToBigIP(method string, route string,
...
+	if err != nil {
+		log.Debug("[as3_log] Reading certificate from configmap error.")
+		certificates = ""
+	} else {
+		for _, v := range cm.Data {
```
ssurenr: Please add a comment to clarify the context.

PR #943, opened 2019-06-17
update verify-interval

kkinch:


kkinch:
If I change the next-version, the page will display 1.9.2, while the release notes display 1.9.1 which is confusing.

https://clouddocs.f5.com/products/connectors/k8s-bigip-ctlr/v1.9/RELEASE-NOTES.html

 I'm not clear what the workaround is for this.

ssurenr:
> If I change the next-version, the page will display 1.9.2, while the release notes display 1.9.1 which is confusing.
> 
> https://clouddocs.f5.com/products/connectors/k8s-bigip-ctlr/v1.9/RELEASE-NOTES.html
> 
> I'm not clear what the workaround is for this.

Next version should point to 1.9.2 at this time. I'll fix it quickly. For the all the patch release the documentation is the same and will be 1.9.

PR #945, opened 2019-06-19
Switch functionality between AS3 and CCCL

srivastavnitin24:
CONTCNTR-755: Implementing a switching functionality to choose between CCCL and AS3


cmd/k8s-bigip-ctlr/main.go
```diff
@@ -108,6 +108,7 @@ var (
 	as3Validation      *bool
 	sslInsecure        *bool
 	trustedCertsCfgmap *string
+	as3Mode            *bool
```
ssurenr: Instead of a boolean, let's use a string input called "agent", which by default set to "cccl". Users can override this to "as3" and later we can set this as a default value. It can accommodate other implementations too.

pkg/appmanager/outputConfig.go
```diff
@@ -27,10 +27,21 @@ import (
 // Dump out the Virtual Server configs to a file
 func (appMgr *Manager) outputConfig() {
 	appMgr.resources.Lock()
-	appMgr.outputConfigLocked()
+	if appMgr.as3Mode {
```
ssurenr:
```suggestion
	if appMgr.agent == "as3" {
```
In accordance with the parameter modification.

pkg/appmanager/outputConfig.go
```diff
@@ -27,10 +27,21 @@ import (
...
+	} else {
+		//CCCL execution
+		appMgr.outputConfigLocked()
+	}
 	appMgr.resources.Unlock()
 }
 
+//TODO : OpenShift routing : AS3 Mode
+func (appMgr *Manager) outputConfigLockedAs3Mode() {
+	log.Debugf("-------------- In As3 Mode : OpenShift Routes Impl --------------", appMgr.as3Mode)
```
ssurenr:
```suggestion
	log.Debugf("Using agent: %v", appMgr.agent)
```

cmd/k8s-bigip-ctlr/main.go
ssurenr: @mdditt2000 Thoughts?

ssurenr:
Add a function to validate `agents`. If value is other than `cccl` or `as3`, log an error and quit the app.

cmd/k8s-bigip-ctlr/main.go
srivastavnitin24: Done the changes as suggested using a string parameter agent

pkg/appmanager/outputConfig.go
srivastavnitin24: Done

pkg/appmanager/outputConfig.go
srivastavnitin24: Done

srivastavnitin24:
Coming up with a new pull request with few changes in the implementation of switch functionality between as3 and cccl

PR #948, opened 2019-06-25
Switch between as3 and cccl using agent parameter

srivastavnitin24:


pkg/appmanager/outputConfig.go
```diff
@@ -18,17 +18,37 @@ package appmanager
...
 	log "github.com/F5Networks/k8s-bigip-ctlr/pkg/vlogger"
 )
 
+// invalidAgent is returned when no matching agent values are given neither "as3" nor "cccl"
+var invalidAgentParam = "Invalid agent parameter provided (agent parameter can only be 'as3' or 'cccl')"
+
 // Dump out the Virtual Server configs to a file
-func (appMgr *Manager) outputConfig() {
+func (appMgr *Manager) outputConfig() error {
 	appMgr.resources.Lock()
```
subbuv26: as opened a lock, add `defer appMgr.resoures.Unlock()` after immediately calling Lock(), instead of calling it before each return.

pkg/appmanager/appManager_test.go
```diff
@@ -583,7 +587,7 @@ var _ = Describe("AppManager Tests", func() {
 			}
 			appMgr := NewManager(&Params{ConfigWriter: mw})
 			Expect(func() { appMgr.outputConfig() }).ToNot(Panic())
-			Expect(mw.WrittenTimes).To(Equal(1))
+			Expect(mw.WrittenTimes).To(Equal(0))
```
trinaths: Please dont modifiy existing unit tests unless there is a certain need. Please share your comments supporting this change.

pkg/appmanager/outputConfig.go
srivastavnitin24: Done changes as suggested

pkg/appmanager/appManager_test.go
srivastavnitin24: Since we have created a fork between cccl and as3 execution the test cases for virtual server need not fail every time since the execution depends upon the agent flag now.
Since the syncVirtualServer func call the outconfig func hence it is not expecting it to fail every time. Therefore it was expecting the arguments as below.
    Expected
        <int>: 0
    to equal
        <int>: 1
We have maintained the rest of the test cases as it is by adding m.appMgr.agent = "cccl" .

PR #963, opened 2019-07-17
Ignore Master Node as Pool member

veeramalla-f5:
Problem:
Master node is considered as a pool member when it is in ready to schedule state.

Solution:
Ignore Master node always.

Affected branches:
master, feature.as3

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -2238,6 +2238,10 @@ func (appMgr *Manager) getNodes(
 
 	// Append list of nodes to watchedNodes
 	for _, node := range nodes {
+		// Ignore Master Node from the list of Watched nodes.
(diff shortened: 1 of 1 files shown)


dgarrison63:
What's the reason for implementing this feature?  What this means is that any nodes with the role of master won't be added as pool members even though customers may taint their masters to allow them to be schedulable.  

Additionally, this will make it impossible to use CIS with minishift, minikube and similar single node cluster deployments.  I often tell customers that are getting started to use minikube, minishift or similar and I use them quite a bit when testing.

My strong suggestion would be to make this a command line argument that customers can set if they don't want to ever use the masters as pool members.

veeramalla-f5:
Thanks for the suggestion David, We didnot think of minikube and minishift usage . However, I will have a discussion with team and revoke the code if needed.

PR #966, opened 2019-07-19
as3 schema validation using latest not version without fall back mechanism

srivastavnitin24:
AS3 schema validation using latest without fall back mechanism.

If the URL is down / not accessible we bypassing the validation.

### DIFF
--- a/pkg/appmanager/as3Manager.go
+++ b/pkg/appmanager/as3Manager.go
@@ -43,6 +43,7 @@ const (
 	svcTenantLabel           = "cis.f5.com/as3-tenant"
 	svcAppLabel              = "cis.f5.com/as3-app"
 	svcPoolLabel             = "cis.f5.com/as3-pool"
+	as3SchemaLatest          = "https://raw.githubusercontent.com/F5Networks/f5-appsvcs-extension/master/schema/latest/as3-schema.json"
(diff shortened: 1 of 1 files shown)


pkg/appmanager/as3Manager.go
```diff
@@ -104,12 +106,34 @@ func (appMgr *Manager) processUserDefinedAS3(template string) bool {
 // Validates the AS3 Template
 func (appMgr *Manager) validateAS3Template(template string) bool {
 
-	var schema = appMgr.schemaLocal + "as3-schema-3.11.0-3-cis.json"
+	res, _ := http.Get(as3SchemaLatest)
```
ssurenr: Let's catch and handle the error.

srivastavnitin24: Handling the error if occurred while fetching the latest.

Named err parameter as resErr to avoid shadowing of variable.

ssurenr:
Please add an entry in [docs/RELEASE-NOTES.rst](https://github.com/github/hubot/blob/master/README.md) under `next release` section

srivastavnitin24:
> Please add an entry in [docs/RELEASE-NOTES.rst](https://github.com/github/hubot/blob/master/README.md) under `next release` section

Done

PR #967, opened 2019-07-23
Created the Gitlab CI.

vklohiya:
Added the GitLab CI with and without glider provisioning.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,125 @@
+stages:
+  - Build
+  - Test
+
+Build-Image:
+  tags: 
+    - docker-executor
+  stage: Build
+  image: ${DOCKER_BUILDER}
+  artifacts:
+    paths:
+      - shared-vars.sh
+  
+  script:
+    - apk update
+    - apk add git make python coreutils bash sudo
+    - if [ "$DOCKER_NAMESPACE" == "" ]; then DOCKER_NAMESPACE="local"; fi
+    - export BASE_PUSH_TARGET="$DOCKER_NAMESPACE/k8s-bigip-ctlr"
+    - |
+      if [ "$CI_REGISTRY_PASSWORD" == "" -o "$CI_REGISTRY_USER" == "" -o $DOCKER_NAMESPACE == "" ]; then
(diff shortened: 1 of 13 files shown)


ssurenr:
@vklohiya Let this PR come to master

build-tools/attributions-generator.sh
```diff
@@ -10,10 +10,15 @@ set -x
 ATTR_GEN_IMG=f5networksdevel/attributions-generator:latest
 docker pull ${ATTR_GEN_IMG}
 
+LOCAL_USER_ID=$(id -u)
+if [ "$GITLAB_CI" == true ]; then
+  LOCAL_USER_ID=9001
```
ssurenr: Why are we hardcoding this?

.gitlab-ci.yml
```diff
@@ -0,0 +1,125 @@
...
+        yum install git unzip wget make openssh python python-pip  -y
+        pip install ansible==2.6.5
+        # Cloning Glider Repo
+        export GIT_ASKPASS=${PWD}/git_config.sh
+        mkdir glider
+        echo $GLIDER_REPO
+        pushd glider
+        echo -e '#!/bin/bash \necho $GIT_PASSWORD' > $GIT_ASKPASS
+        chmod +x $GIT_ASKPASS
+        git init && git config http.sslVerify false
```
ssurenr: Let's either use proper CA certificates to trust or use SSH Keys for authentication.

.gitlab-ci.yml
```diff
@@ -0,0 +1,125 @@
...
+        git config credential.https://gitswarm.f5net.com.username ${GITLAB_USER_LOGIN}
```
ssurenr: `gitswarm` is deprecated. Let's use `gitlab.f5net.com`

build-tools/attributions-generator.sh
vklohiya: As GITLAB container does not have any other user created we are using the root user, however we need some id for CIS container to run, hence using 9001, earlier it was using the ID of user which is running the make prod command.

ssurenr:
Tracked in #980

PR #987, opened 2019-08-06
Add openshift_passthrough_irule for edge routes

veeramalla-f5:
Problem:
default-route-serverssl profile may not get removed when Reencrypt routes are modified to edge.

Solution:
Add openshift_passthrough_irule to virtual server and update the ssl_reencrypt_serverssl_dg
with key as hostname and value empty, This will result in forwarding http traffic to backend(service).

Affected Branches:
master


ssurenr:
Please add an entry in [docs/RELEASE-NOTES.rst](https://github.com/github/hubot/blob/master/README.md) under `next release` section

veeramalla-f5:
Handled as part of https://github.com/F5Networks/k8s-bigip-ctlr/pull/990

PR #990, opened 2019-08-08
Handle the delete of default-route-serverssl when there are no reencr…

veeramalla-f5:
Problem:
Handle the delete of default-route-serverssl when there are no reencrypt routes

Solution:
Get the list of routes and check if there are any reencrypt routes. If there are no reencrypt routes delete default-route-serverssl profile as it is not used by any virtual server.

Affected branches:
master

### DIFF
--- a/pkg/appmanager/profiles.go
+++ b/pkg/appmanager/profiles.go
@@ -26,7 +26,7 @@ import (
 	log "github.com/F5Networks/k8s-bigip-ctlr/pkg/vlogger"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/pkg/api/v1"
+	v1 "k8s.io/client-go/pkg/api/v1"
 	"k8s.io/client-go/pkg/apis/extensions/v1beta1"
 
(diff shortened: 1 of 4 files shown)


pkg/appmanager/profiles.go
```diff
@@ -426,9 +440,15 @@ func (appMgr *Manager) deleteUnusedProfiles(
...
+
+			// If reencryptRouteCount >= 1, don't remove default-route-serverssl
+			if reencryptRouteCount >= 1 && prof.Name == "default-route-serverssl" {
+				reencryptRouteCount = 0
+				continue
+			}
 			// Don't process profiles that came from a resource in a different namespace.
 			// We don't want to delete them, since they won't reference a resource this time.
-			if prof.Namespace != namespace {
+			if prof.Namespace != namespace && prof.Name != "default-route-serverssl" {
```
trinaths: Since we don't care about resources in other NS, why should you check the name ?

veeramalla-f5: This is needed. This method iterates over all the profiles, since the default profiles does not have any namespaces, these profiles are ignored for deletion.

trinaths:
Looks good to me

PR #991, opened 2019-08-12
Start monitoring of pods on Unscheduled nodes.

veeramalla-f5:
Problem:
CIS does not monitor the pods on unscheduled nodes. This results in traffic loss.

Solution:
Start monitoring of pods on Unscheduled nodes. Ignore master node only when it is in unscheduled state. This will avoid the load on kube-proxy inside the master node. 

Affected Branches:
master


ssurenr:
We need to add all pool members regardless of the node status. For now, this can be taken as mitigation.

PR #995, opened 2019-08-16
Implement Routes using AS3 orchestration

subbuv26:
Provide user to choose between AS3 orchestration or default to CCCL orchestration to use.
If AS3 orchestration is used, use <partition>_AS3 for LTM objects and <partition> for Network Objects.
Generate AS3 JSON stanza from Openshift Routes.
Combine User-defined AS3 JSON and AS3 JSON and post to BIG-IP.


pkg/appmanager/as3Manager.go
```diff
@@ -687,9 +698,533 @@ func (appMgr *Manager) checkValidAS3Endpoints(obj interface{}) (
...
+				if val, ok := svc.ClientTLS.(string); ok {
+					return strings.Join([]string{"", DEFAULT_PARTITION, as3SharedApplication, val}, "/"), true
+				}
+				log.Errorf("Unable to find serverssl for Data Group: %v\n", dgName)
+			}
+		}
+	}
+	return "", false
+}
+
```
trinaths: as a practice, we need to have comments about the functions for readability and understanding what really it does.

subbuv26: This needs to be taken as part of refactoring.

PR #1010, opened 2019-08-29
Add Debugger Support

ssurenr:
Problem: It is not possibe to debug controller from IDE using debugger
such as Go Delve.

Solution: Introduced a target to Makefile named `debug` which will
disable compiler optimization and embed *dlv* binary in to the
controller image.

Signed-off-by: Surendhar <surendhar@ymail.com>


ssurenr:
A sample image generated by this modification is available in this tag: `ssurenr/k8s-bigip-ctlr:debug`

PR #1026, opened 2019-09-25
route-vserver-addr is required when mange-routes is enabled

trinaths:
Problem: When manage-routes is enabled and route-vserver-addr is not given then deployment fails with unknown error.

Solution: Check whether cis deployment option route-vserver-addr is also given when manage-routes is enabled. This will not crash the deployment but provide a detailed help message.

Affected-branches: master


trinaths:
Regression completed. No issues.

PR #1029, opened 2019-09-26
as3 schema validation using latest schema

srivastavnitin24:
Problem:
Currently, AS3 schema validation uses a particular version to validate the schema and not the latest

Solution:
This fix will allow AS3 schema validation to be done with the latest/updated AS3 schema 

Affected Branches:
master

### DIFF
--- a/cmd/k8s-bigip-ctlr/main.go
+++ b/cmd/k8s-bigip-ctlr/main.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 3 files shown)


cmd/k8s-bigip-ctlr/main.go
```diff
@@ -733,6 +736,9 @@ func main() {
 		appMgr.DeleteAS3ManagedPartition()
 	}
 
+	// AS3 schema validation using latest AS3 version
+	as3SchemaValidation(appMgr)
```
ssurenr: Rename function to `fetchAS3Schema`

srivastavnitin24: Done

PR #1032, opened 2019-09-27
Update Route admitted status in OpenShift 3.11 GUI

trinaths:
Problem: In Openshift ROUTE status is updated by its default router. We need to update it with respect to BIG-IP acceptance of the route.

Solution: OpenShift default router is scaled down. BIG-IP controller is updated to handle Route ADMITTED status based on the response from BIGIP. 

Affected-branches: master

Signed-off-by: Trinath SOMANCHI trinath.somanchi@gmail.com

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -205,6 +205,8 @@ type RouteConfig struct {
 	ServerSSL   string
 }
 
+var RoutesProcessed []*routeapi.Route
+
 // Create and return a new app manager that meets the Manager interface
 func NewManager(params *Params) *Manager {
 	vsQueue := workqueue.NewNamedRateLimitingQueue(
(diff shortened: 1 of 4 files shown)


pkg/appmanager/appManager.go
```diff
@@ -1361,6 +1363,9 @@ func (appMgr *Manager) syncRoutes(
 	for _, route := range routeByIndex {
 		if route.ObjectMeta.Namespace != sKey.Namespace {
 			continue
+		} else {
```
ssurenr: else block is not required. keep the logic outside if condition.

trinaths: ok.

ssurenr:
OSCP 4.1 have problems updating route status.

trinaths:
This fix addresses GUI status update for OpenShift 3.11 only.

PR #1039, opened 2019-10-10
Fix #793: CC does not deal with the situation where one URI for a hos…

### LINKED ISSUE #793: iRule used to select server-ssl profile does not take into account URI path routing.
#### Description
In OpenShift, routes can have different TLS termination configurations within a single host-name.

For example, consider the following routes:
```
oc get routes
NAME                   HOST/PORT          PATH          SERVICES        PORT          TERMINATION          WILDCARD
example.com-static     example.com        /             crs-app-anf     https         reencrypt            None
example.com            example.com        /static       crs-app-anf     http-static   edge                 None
```

Everything for example.com should be re-encrypt with a server-ssl profile, EXCEPT /static, which should be edgeTerminated and be sent to the container in cleartext.

Right now, the decision to re-encrypt and apply a server-profile happens in the iRule at the ClientSSL handshake.  This is too early to support the fact that that some paths may need re-encrypted while others may not.

Currently, k8s-bigip-ctlr will apply a server-ssl profile for any request to example.com, re-encrypting all paths.  Ideally, this decision would be made later in the process at HTTP_REQUEST so that the URI could be taken into account before applying a server-ssl profile.

#### OpenShift Version
3.11

#### Controller Version
1.7

#### BIG-IP Version
13.1.1.0.8

#### Diagnostic Information

In general this issue is due to the fact that the decision to enable server-ssl profile, and which server-ssl profile to apply, should happen after the HTTP request is received on the client-side, and before it being sent to the server-side.  Currently, it happens either too early in the process, and/or does not take into account the HTTP::path of the request from the client.

Currently, the decision to enable server-ssl is made in the openshift_passthrough_irule, at line 126 though 131:
```
elseif { [class exists $reencrypt_class] } {
	set dflt_pool [class match -value $servername_lower equals $reencrypt_class]
	if { not ($dflt_pool equals "") } {
		SSL::enable serverside
	}
}
```
There are two issues here:
1) The data-group only has host-name data, no URI data for an iRule to make a decision.
2) This happens in the CLIENT_DATA, event, and we only get into this logic block if the packet received is a clientssl handshake.

A similar issue occurs when the iRule selects which server-ssl profile should be used.  Currently, this happens in SERVER_CONNECTED which is fine, however it does not take the HTTP::path from the client-side of the reuqest into account when making the decision.  The data-groups need to contain both server name & path, then the logic should be modified to inspect the path and select a server-ssl profile that is appropriate.
```
when SERVER_CONNECTED {
	set svrssl_class "/openshift-ocpd-dmz/ssl_reencrypt_serverssl_dg"
	if { [class exists $svrssl_class] } {
		set profile [class match -value $servername_lower equals $svrssl_class]
		if { not ($profile equals "") } {
			SSL::profile $profile
		}
	}
}
```

veeramalla-f5:
…tname should be re-encrypted and another URI should be edge terminated.

Problem: CC does not deal with the situation where one URI for a hostname should be re-encrypted and another URI should be edge terminated

Solution: 
As of today Server SSL Datagroup(ReencryptServersslDg) created by CIS creates keyvalue pair with key as ServerName and value as ServerSSL. When a reencrypt route and Edge route are attached to the same hostname, the ServerSSL of reencrypt route is used by Edge Route as well.
The best way to overcome this situation is to update the Datagroup with Servername and Path.
Example:
For Route With Host-> foo.com and Path-> /foo ......... update the data group key as foo.com/foo instead of foo.com

Image: abhishekf5/k8s-bigip-ctlr:updatedatagroup

Affected Branches: master

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -145,7 +145,7 @@ type Manager struct {
 	WatchedNS       WatchedNamespaces
 	as3RouteCfg     ActiveAS3Route
 	As3SchemaLatest string
-	intF5Res        InternalF5Resources // AS3 Specific features that can be applied to a Route/Ingress
+	intF5Res        InternalF5ResourcesGroup // AS3 Specific features that can be applied to a Route/Ingress
 }
 
 // FIXME: Refactor to have one struct to hold all AS3 specific data.
@@ -248,6 +248,7 @@ func NewManager(params *Params) *Manager {
 		sslInsecure:        params.SSLInsecure,
 		trustedCertsCfgmap: params.TrustedCertsCfgmap,
 		Agent:              getValidAgent(params.Agent),
+		intF5Res:           make(map[string]InternalF5Resources),
 	}
 	if nil != manager.kubeClient && nil == manager.restClientv1 {
 		// This is the normal production case, but need the checks for unit tests.
@@ -1364,7 +1365,7 @@ func (appMgr *Manager) syncRoutes(
 	svcFwdRulesMap := NewServiceFwdRuleMap()
 
 	// buffer to hold F5Resources till all routes are processed
-	bufferF5Res := map[Record]F5Resources{}
+	bufferF5Res := InternalF5Resources{}
 
 	for _, route := range routeByIndex {
 		if route.ObjectMeta.Namespace != sKey.Namespace {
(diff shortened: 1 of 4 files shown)


pkg/appmanager/appManager.go
```diff
@@ -1437,8 +1437,19 @@ func (appMgr *Manager) syncRoutes(
 					appMgr.setClientSslProfile(stats, sKey, rsCfg, route)
 					serverSsl := appMgr.setServerSslProfile(stats, sKey, rsCfg, route)
 					if "" != serverSsl {
+						// Combination of hostName and path are used as key in reencryptdssl Datagroup.
```
trinaths: re-encryptssl

trinaths: comments typo.

veeramalla-f5: Thanks Trinath. Fixed it.

PR #1044, opened 2019-10-14
OCP 3.11 and 4.1 - Fix for Route Admit statuses.

trinaths:
Problem: The status update in OCP 3.11 was done while scale-down of openshift-default-router. But this is not possible with OCP 4.x and higher versions where a OpenShift default router always exists.

Solution: Check where there is an update from F5 BIG-IP router, else update the condition. This makes F5 route status update co-exists with openshift default router updates.

Signed-off-by: Trinath Somanchi <trinath.somanchi@gmail.com>


nandakishorepeddi:
@trinaths @ssurenr With this fix, CIS currently doesnt update admitted status of the route when modified

trinaths:
Raising a new PR

PR #1052, opened 2019-10-22
CIS will no more generate dozens of log messages every 30 seconds

snatra:
Problem: CIS 1.11 generating dozens of "INFO" log messages every 30 seconds.

Solution: Update operation of "Internal F5 Resource" not handled correctly, an update should not 
               happen when both Internal F5 Resource and Buffer F5 Resource are empty. fixed this by 
               adding a condition

               Also, this pull request contains a fix for handling "outputConfig" appropriately on agent= 
               As3.  CIS must not use "outputConfigLocked" method when configured --agent=AS3.
               Made changes so that "outputConfig" method only will be called when configured  
               Agent as AS3.

Build : snatra27/k8s-bigip-ctlr-logger:1.0

Affected-branches: Master, 1.11-stable


ssurenr:
Please rebase and squash the commits. Release notes is coming as seperate entity.

PR #1054, opened 2019-10-23
Fix applying WAF to root of a domain

subbuv26:
Problem: WAF Policy is not being applied in virtual if Path in the OCP route is "/"

Solution: Remove ending "/" in a path to fix the issue.

Affected-Branches: master, 1.11-stable


subbuv26:
Docker Image: subbuv26/k8s-bigip-ctlr:waf-path-fix

PR #1062, opened 2019-10-29
Fix calling unsafer method of Output Config

snatra:
Problem: outputConfigLocked is being called without locking properly

Solution: Call outputConfig instead of outputConfigLocked so that locking of resources is ensured


snatra:
docker-image: snatra27/k8s-bigip-ctrl:lock-fix

PR #1064, opened 2019-10-30
Fix Controleller crash with WAF Route and unavaible service

subbuv26:
Problem: Controller is crashing when Route with WAF Policy is deployed with unavailable service

Solution: Make sure, policy exists in AS3 declaration before applying WAF related to conditions and actions. Because Controller does not create policy if Route is configured with unavailable service.


subbuv26:
Docker Image: subbuv26/k8s-bigip-ctlr:1.11.1_build_3

PR #1072, opened 2019-11-02
In CCCL mode with AS3 CfgMap, CIS continuously logs "[INFO] Wrote 0 Virtual Server and 0 IApp configs"

snatra:
Problem: CIS continuously logs this message "[INFO] Wrote 0 Virtual Server and 0 IApp configs"

Solution: Properly handled the AS3 members in CIS

affects-branches: master

docker-image: snatra27/k8s-bigip-ctlr:rel12_build_5

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -11,7 +11,12 @@ Bug Fixes
 `````````
 * Controller handles combination of Edge and Reencrypt Openshift routes.
 * Controller now does not send encrypted traffic to backend.
+* Controller now does not log dozens of "INFO" log messages frequently in CCCL mode.
+* Ingress in K8S cluster mode, observing no data plane traffic as there were no static ARP entries available in BIG-IP.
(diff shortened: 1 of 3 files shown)


ssurenr:
@snatra This is a known issue since 1.9 release. This should not cause any issues. This needs a proper regression testing and manual testing. Why are we looking at this to 1.11.1 patch? It is safe to fix it in a minor release.

snatra:
> @snatra This is a known issue since 1.9 release. This should not cause any issues. This needs a proper regression testing and manual testing. Why are we looking at this to 1.11.1 patch? It is safe to fix it in a minor release.

This is producing a dump of "[INFO] Wrote 0 Virtual Server and 0 IApp configs" in CCCL for configmap cases
This issue is what we are releasing in 1.11.1.

ssurenr:
@snatra Since this bug is not introduced after 1.11.0, let's fix it in 1.12 minor release.

PR #1075, opened 2019-11-04
Migrate from godeps to go modules

veeramalla-f5:
Problem:
Upgrade to Go modules

Solution:
Migrate to Go mods from Godeps using the go mod init and go mod tidy


trinaths:
To pass this at CI, we need to update the attributions generator.

PR #1076, opened 2019-11-06
Added new command line option, --log-response-body which will log the…

agupta49:
Problem: Controller logs the response from AS3 in BIGIP as is, which sometimes include the original request in plain text. As a result, if the AS3 request consists of secrets such as credentials, passphrase, private keys etc are getting logged in plain text.

Solution: Given the option to hide AS3 response body.
Introduce a new command-line option to k8s-bigip-ctlr Binay, named --log-as3-response which will log the AS3 response body in Controller logs. This is boolean parameter that accepts either true or false as input. By default this option should be set to false.

Affects-branches: master

docker-image: amit49g/k8s-bigip-ctlr:logResponseBodyOption

### DIFF
--- a/cmd/k8s-bigip-ctlr/main.go
+++ b/cmd/k8s-bigip-ctlr/main.go
@@ -112,6 +112,7 @@ var (
 	sslInsecure        *bool
 	trustedCertsCfgmap *string
 	agent              *string
+	logAS3Response     *bool
 
 	vxlanMode        string
 	openshiftSDNName *string
@@ -183,6 +184,8 @@ func _init() {
 		"Optional, when set to false, disables as3 template validation on the controller.")
 	sslInsecure = bigIPFlags.Bool("insecure", false,
 		"Optional, when set to true, enable insecure SSL communication to BIGIP.")
+	logAS3Response = bigIPFlags.Bool("log-as3-response", false,
+		"Optional, when set to true, add the body of AS3 API response in Controller logs.")
 	trustedCertsCfgmap = bigIPFlags.String("trusted-certs-cfgmap", "",
 		"Optional, when certificates are provided, adds them to controller’s trusted certificate store.")
 	// TODO: Rephrase agent functionality
@@ -651,6 +654,7 @@ func main() {
 		SSLInsecure:        *sslInsecure,
 		TrustedCertsCfgmap: *trustedCertsCfgmap,
 		Agent:              *agent,
+		LogAS3Response:     *logAS3Response,
 	}
 
 	// If running with Flannel, create an event channel that the appManager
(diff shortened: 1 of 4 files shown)


pkg/appmanager/as3Manager.go
```diff
@@ -574,8 +574,10 @@ func (as3RestClient *AS3RESTClient) restCallToBigIP(method string, route string,
 			log.Debugf("[as3_log] Big-IP Response code: %v,Response:%v, Message: %v", v["code"], v["response"], v["message"])
 		}
 	} else {
-		//log.Debugf("[as3_log] Big-IP Response error: %v", response) // need fix to dump only non sensitive data
 		log.Debugf("[as3_log] Big-IP Responded with error code: %v", resp.StatusCode)
+		if appMgr.logResponseBody {
+			log.Debugf("[as3_log] Raw response from Big-IP: %v ", response)
```
snatra: Please Change [as3_log] to [AS3] in log

cmd/k8s-bigip-ctlr/main.go
```diff
@@ -183,6 +184,8 @@ func _init() {
 		"Optional, when set to false, disables as3 template validation on the controller.")
 	sslInsecure = bigIPFlags.Bool("insecure", false,
 		"Optional, when set to true, enable insecure SSL communication to BIGIP.")
+	logResponseBody = bigIPFlags.Bool("log-response-body", false,
+		"Optional, when set to true, enable log the AS3 response body in Controller logs.")
```
ssurenr:
```suggestion
		"Optional, when set to true, add the body of AS3 API calls in Controller logs.")
```

pkg/appmanager/as3Manager.go
agupta49: Yes. It's changed.

cmd/k8s-bigip-ctlr/main.go
```diff
@@ -112,6 +112,7 @@ var (
 	sslInsecure        *bool
 	trustedCertsCfgmap *string
 	agent              *string
+	logResponseBody    *bool
```
trinaths: logResponseBody - the option "log-response-body" and this param are a bit ambiguous.

Can we have something like, log-bigip-response ??

cmd/k8s-bigip-ctlr/main.go
```diff
@@ -183,6 +184,8 @@ func _init() {
 		"Optional, when set to false, disables as3 template validation on the controller.")
 	sslInsecure = bigIPFlags.Bool("insecure", false,
 		"Optional, when set to true, enable insecure SSL communication to BIGIP.")
+	logResponseBody = bigIPFlags.Bool("log-response-body", false,
+		"Optional, when set to true, add the body of AS3 API calls in Controller logs.")
```
trinaths: is that, AS3 API call or response ?

agupta49: This is only for AS3 API call.

cmd/k8s-bigip-ctlr/main.go
trinaths: log-as3-response

cmd/k8s-bigip-ctlr/main.go
trinaths: Optional, when set to true, add the body of AS3 API response in Controller logs.

cmd/k8s-bigip-ctlr/main.go
agupta49: Changed to log-as3-response.

PR #1106, opened 2019-11-29
Update secure-serverssl behaviour

subbuv26:
Requirement: secure-serverssl is applied across ssl profile in AS3, so even if one route has secure-serverssl enabled, all the hosts that share same ssl profile will be validated with corresponding certificates.

Problem: With previous implementation setting of secure-serverssl depends on the order of processing the routes

Solution: Even if one route has secure-serverssl, all the certificates of hosts that share the same ssl profile will be validated


subbuv26:
Docker Image: subbuv26/k8s-bigip-ctlr:rel_build_11

PR #1124, opened 2019-12-16
Data group inconsistency issue fix

agupta49:
Problem: Data group was inconsistent when routes are created in multiple namespaces.

Solution: Fixed the issue while generating AS3 declaration for DG.

Image: amit49g/k8s-bigip-ctlr:dgFix


ssurenr:
@agupta49 Please add a point in the release notes.

agupta49:
> @agupta49 Please add a point in the release notes.

I added this on bug fixes section.

PR #1129, opened 2019-12-17
Support Ingress.Class and Manage default Ingresses.

### LINKED ISSUE #976: Ingress Resource - Only handle annotated ingresses
#### Description
It seems that the F5 BIG-IP Controller watches for all `Ingress` resources that either have `kubernetes.io/ingress.class: f5` or that don't have the annotation at all. In a cluster that has multiple ingress controllers, it ends up with unpredictable results for ingresses that do not specify their class, the different controllers fighting against each other to handle the resource.

I would propose to add an option to the controller so that it does not try to handle ingresses that do not have the class annotation, e.g. `--manage-default-ingresses=false`, or one that explicitly defines the class to watch, e.g. `--ingress-class=f5`, or both.

#### Kubernetes Version
1.14

#### Controller Version
1.9

trinaths:
**Problem:**

[1] CIS watches all `Ingress` resources that either have `kubernetes.io/ingress.class: f5` or that don't have that annotation at all.
In a cluster that has multiple ingress controllers, it ends up with unpredictable results for ingresses that do not specify their class, the different controllers fighting against each other to handle the resource.

[2] CIS processes ingresses with ingress.class defaults to "f5" and there is no option for custom ingress classes.

**Solution:**

New controller deployment options:

- `--manage-ingress-class-only` (default - `false`) is added. when set to `true`, only considers `ingress.class` annotated ingresses. Using the below option, user can configure the controller with custom `ingress.class`. 
-  `--ingress-class` (default - `f5`) is added to enable CIS to process Ingresses with custom ingress.class.

**Build**: somanchit/k8s-bigip-ctlr:ingclass_fix

**Affected Branches**: master

**Release:** 1.13

Fixes #976

### DIFF
--- a/pkg/appmanager/resourceConfig.go
+++ b/pkg/appmanager/resourceConfig.go
@@ -1451,11 +1451,16 @@ func (appMgr *Manager) createRSConfigFromIngress(
 	snatPoolName string,
 ) *ResourceConfig {
 	if class, ok := ing.ObjectMeta.Annotations[k8sIngressClass]; ok == true {
-		if class != "f5" {
+		if class != appMgr.ingressClass {
+			return nil
+		}
+	} else {
+		// at this point we dont have k8sIngressClass defined in Ingress definition.
+		// So check whether we need to process those ingress or not.
+		if appMgr.manageIngressClassOnly {
 			return nil
 		}
 	}
-
 	var cfg ResourceConfig
 	var balance string
 	if bal, ok := ing.ObjectMeta.Annotations[f5VsBalanceAnnotation]; ok == true {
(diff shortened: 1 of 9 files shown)


veeramalla-f5:
Can we change the name from --manage-default-ingress to --manage-all-ingress

Regards,
AV

trinaths:
> Can we change the name from --manage-default-ingress to --manage-all-ingress
> 
> Regards,
> AV

Sure. We can.

pkg/appmanager/resourceConfig.go
```diff
@@ -1450,12 +1450,18 @@ func (appMgr *Manager) createRSConfigFromIngress(
 	defaultIP,
 	snatPoolName string,
 ) *ResourceConfig {
+	// if manage-default-ingress is false, then process ingress with ingress.class=f5 annotation only
```
snatra: Please change in comment "manage-default-ingress" to "manage-ingress-class-only"

trinaths: sure.

Issue #1014, opened 2019-09-03
Ingress resource - Performance problem with large number of ingresses


codertao:
#### Description
We're having a performance issue with the k8s-bigip-ctlr ; we have ~30 ingresses with unique certs being served off a single virtual server. 

I *think*, due to: https://github.com/F5Networks/k8s-bigip-ctlr/blob/11f07d9dfa6957813ac88be8a4701da7744bce67/pkg/appmanager/resourceConfig.go#L1677 , each call to appManager.updateVirtualServer pulls down, as far as I can tell, every secret used (by the virtual server?), which is taking 200ms a call. Net result is ~4 seconds for a call to updateVirtualServer, and 2-5 minutes on average for a new pod in a service to be added to the related pool. Or 2-5 minutes of downtime for rolling deployments.

#### Kubernetes Version
1.12, Amazon EKS

#### Controller Version
1.9, built in house

#### BIG-IP Version
14.0.0.1

#### Diagnostic Information
Quite awful custom logging added to processNextVirtualServer:

2019/09/03 21:32:06 [INFO] Process next virtual server some-app, queue: 44
2019/09/03 21:32:06 [INFO] Start lookup service {Namespace:staging ServiceName:some-app As3Name: As3Data:} (1.882µs)
2019/09/03 21:32:06 [INFO] Get resources for Key {Namespace:staging ServiceName:some-app As3Name: As3Data:} (26.63µs)
2019/09/03 21:32:06 [INFO] sync configmap {Namespace:staging ServiceName:some-app As3Name: As3Data:} (36.167µs)
2019/09/03 21:32:06 [INFO] sync ingress {Namespace:staging ServiceName:some-app As3Name: As3Data:} (12.498163ms)
2019/09/03 21:32:06 [INFO] . . . . . Startinng ing rule handling 402ns - 11.248µs
2019/09/03 21:32:06 [INFO] . . . . . Finished dependency stuff 16.859µs - 27.708µs
2019/09/03 21:32:06 [INFO] . . . . . . Created RSConfig 103.711µs - 114.608µs
2019/09/03 21:32:06 [INFO] . . . . . . Handled TLS 183.947µs - 194.845µs
2019/09/03 21:32:06 [INFO] . . . . . . Handled Health 222.576µs - 233.454µs
2019/09/03 21:32:06 [INFO] . . . . . . Handled Something Fast 276.78µs - 287.698µs
[...]

PR #1133, opened 2019-12-19
Performance issue in CIS when simultaneously processing more number of ingress resources

snatra:
Problem: Ingress resource - Performance problem with a large number of ingresses(git hub issue #1014)

Solution: Optimized the Kubernetes client API calls in CIS while processing Ingress resources

Docker Image: snatra27/k8s-bigip-ctlr:ingress-performance-issue


veeramalla-f5:
Looks good.

PR #1139, opened 2019-12-31
Log json unmarshall failed error body

agupta49:
Problem: 1. Low visibility of AS3 response error due to JSON unmarshal failed.
               2. Override issue with default http and tcp profile of virtual server
                
Solution: Logging AS3 response error body.
                Removing of default HTTP and TCP profile from the virtual server.

Affected branch: master


pkg/appmanager/as3Manager.go
```diff
@@ -1334,19 +1337,6 @@ func createServiceDecl(cfg *ResourceConfig, sharedApp as3Application) {
 
 	svc.Class = "Service_HTTP"
 
-	for _, prof := range cfg.Virtual.Profiles {
-		switch prof.Name {
-		case "http":
-			svc.ProfileHTTP = as3ResourcePointer{
```
veeramalla-f5: I think its not a good idea to remove the http profile from virtual server. Can we look for an alternative?

PR #1140, opened 2020-01-02
Log unmarshal failed error body and remove default http and tcp profi…

agupta49:
Problem: 1. Low visibility of AS3 response error due to JSON unmarshal failed.
2. Override issue with default http and tcp profile of virtual server

Solution: Logging AS3 response error body.
Removing of default HTTP and TCP profile from AS3 declaration (virtual server).

Affected branch: master


trinaths:
please also update release notes.

PR #1178, opened 2020-02-12
Fix ssl redirect when ingress has no rules and host path

vincentmli:
Problem:

when ingress has no rules, host, path as below:

apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: f5ingress
  annotations:
     nginx.ingress.kubernetes.io/rewrite-target: /
     ingress.kubernetes.io/allow-http: "false"
     ingress.kubernetes.io/allow-https: "true"
     kubernetes.io/ingress.class: f5
     ingress.kubernetes.io/ssl-redirect: "true"
     'virtual-server.f5.com/ip': '10.169.72.163'
spec:
  tls:
  - secretName: Common/clientssl
  backend:
    serviceName: nginxservice
    servicePort: 80

SSL redirect irule would fail because SSL redirect irule needs to
match path in datagroup to redirect HTTP to HTTPs, but ingress above
has no rules,host, path defined. Users prefer doing wildcard SSL
redirect without HTTP path matching

Solution:

apply wildcard SSL redirect when ingress has no rules, host, path

### DIFF
--- a/pkg/appmanager/resourceConfig.go
+++ b/pkg/appmanager/resourceConfig.go
@@ -1678,6 +1678,10 @@ func (appMgr *Manager) handleIngressTls(
 	}
 
 	var httpsPort int32
+	var httpPath bool
+	if len(ing.Spec.Rules) > 0 {
+		httpPath = true
+	}
 	if port, ok :=
 		ing.ObjectMeta.Annotations[f5VsHttpsPortAnnotation]; ok == true {
 		p, _ := strconv.ParseInt(port, 10, 32)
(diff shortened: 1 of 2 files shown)


veeramalla-f5:
This does not work fine when there are 2 ingress (1 with ingress.spec.rules and other without ingress.spec.rules) pointing to the same virtual server. It works perfectly when they use two different virtuals on BIG-IP.

trinaths:
Closing this PR due to inactivity.

PR #1183, opened 2020-02-18
Fix Edge Route Redirect issue with WAF

subbuv26:
Problem: Edge Redirect Route is not working with combination of Edge Allow Route or insecure Route.

Root Cause: Edge Redirect route doesn't have a policy in http virtual, which is needed by WAF to decide to drop or allow a request to virtual.

Solution: Add a policy to http virtual, in case of Edge Redirect Route, so that the policy will be consumed by WAF later, so that traffic reaches http virtual and eventually gets redirected to https virtual


subbuv26:
Docker Image: subbuv26/k8s-bigip-ctlr:edge_redirect_fix

PR #1223, opened 2020-03-30
Handle requests sent to a host that is not configured on F5 OpenShift Virtual Server.

veeramalla-f5:
Problem:
CIS throws an error from Irule when user sends a request to a host that is not configured on F5 OpenShift Virtual Server.

User has routes as shown below:
```
NAME    HOST/PORT    PATH  SERVICES TERMINATION   WILDCARD
e1     example.com  svc-1  <all>    reencrypt         None
```
User tries to access the F5 OpenShift Virtual Server with host foo.com, CIS throws Irule error as shown below.

```
TCL error: /example_AS3/Shared/openshift_passthrough_irule <CLIENTSSL_DATA> - can't read "dflt_pool": no such variable     while executing "if { $dflt_pool == "" } then {    log local0.debug "Unable to find pool for $servername_lower"                     } else {      ..."
```

Instead, CIS should smoothly handle such requests instead of throwing Irule error. Irule should pass over such requests to F5 WAF.

Solution:
Handle requests sent to unknown hosts for Routes using debug messages. Users can look at such requests in F5 BIG-IP /var/log/ltm file.

Target Branch:
Master

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -3,6 +3,10 @@ Release Notes for BIG-IP Controller for Kubernetes
 
 Next Release
 ------------
+Bug Fixes
+`````````
+* Controller handles requests sent to a host that is not configured on F5 Virtual Server for Openshift Routes.
+    - Controller will stop throwing Irule TCL errors when user sends requests to a host that is not configured on F5 Virtual Server.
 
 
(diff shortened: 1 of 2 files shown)


trinaths:
@veeramalla-f5  - Code is accepted to agent code 2.0. Please close this PR.

PR #1227, opened 2020-04-02
CIS supports for TEEMS from AS3

trinaths:
This fix is valid for AS3 >= 3.18.


trinaths:
integrated to 2.0 agent code.

PR #1296, opened 2020-05-13
[DOCS] Deploying CIS on OpenShift and HA setup YAML

trinaths:
Required YAML files for reference in docs to deploy
CIS on OpenShift multi node and HA setup.


trinaths:
@veeramalla-f5 - For HA we have a section in docs. also, HA configs are separate from simple deployments.

PR #1297, opened 2020-05-13
Fix for make dev that enable AS3 declaration logging for developer image

agupta49:
"make dev" image build that enable AS3 declaration logging was failing due to outdated patch.
Created new patch to fix this issue.


agupta49:
> Change looks good to me. This needs to be taken care whenever there is a change the code organization.

Sure, Trinath.

PR #1466, opened 2020-08-18
[CRD]: Fix issues with Data Group Creation

iam-veeramalla:
Problem:
CIS is creating DataGroup records before custom resource is validated.

Solution:
DataGroup records should be only created once the Custom Resource is validated.

- Abhishek Veeramalla


agupta49:
Squash the commit into one.

subbuv26 (approved):
Looks Good To Me

PR #1487, opened 2020-08-28
Added support for SNAT in Virtual Server CRD

vklohiya:


pkg/crmanager/resourceConfig.go
```diff
@@ -387,6 +389,12 @@ func (crMgr *CRManager) prepareRSConfigFromVirtualServer(
 	rsCfg.Pools = append(rsCfg.Pools, pools...)
 	rsCfg.Monitors = append(rsCfg.Monitors, monitors...)
 
+	// set the SNAT policy to auto is it's not defined by end user
+	if !(len(vs.Spec.SNAT) > 0) {
```
agupta49, line 393: This should be vs.Spec.SNAT==""

vklohiya: Update the code.

pkg/crmanager/backend.go
```diff
@@ -386,6 +386,13 @@ func createServiceDecl(cfg *ResourceConfig, sharedApp as3Application) {
 	svc.TranslateServerAddress = true
 	svc.TranslateServerPort = true
 
+	if cfg.Virtual.SNAT == "auto" || cfg.Virtual.SNAT =="none" {
```
agupta49, line 389: This code block is not formatted properly.

vklohiya: Addressed formatting issue

agupta49 (approved):

PR #1511, opened 2020-09-18
Fix issues with reading TLSProfile Intermittently

veeramalla-f5:
Hi, 

 After a lot of research, we can now conclude that the issue is with the kubernetes code-generator package. However, as per few recommendations made in the community, I have made the below change to CIS which did not eliminate the issue but reduced the frequency of the issue.

Change in CIS ---> Instead of reading the namespace from the ObjectMeta of the TLSProfile Custom Resource... Pass the namespace of the VirtualServer Custom Resource as an argument to function which is trying to read the TLSProfile. 

I have raised the below issue on kubernetes organisation, code-generator repository in GitHub.
https://github.com/kubernetes/code-generator/issues/116

Regards,
Abhishek Veeramalla

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -1,6 +1,18 @@
 Release Notes for Container Ingress Services for Kubernetes & OpenShift
 =======================================================================
 
+2.2.0
+-------------
+Added Functionality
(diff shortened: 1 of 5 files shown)


agupta49 (approved):
LGTM

PR #1513, opened 2020-09-21
[CRD] Added WAF support

agupta49:
Added WAF support in CRD.
Added doc and example.

Signed-off-by: Amit Kumar Gupta <amitg2812@gmail.com>


veeramalla-f5 (approved):

jmcalalang:
there is nothing in the docs for an example, can this be called out?

There are currently 3 different ways i know to reference a policy and this only accepts a policy existing on a big-ip.

These are the three different kinds i know about:
- on bigip
- at URL endpoing
- from APPolicy crd policy (NGINX)

PR #1563, opened 2020-11-06
Added support for multiport services in config-map

vklohiya:
Added support for multiport services in config-map


vklohiya:
Travis is failing hence closing it.

PR #1564, opened 2020-11-09
Added support for multiport services in config-map

vklohiya:
Added support for multiport services in config-map


subbuv26 (approved):
Looks Good.

PR #1569, opened 2020-11-12
ncc passthrough class update

lavanya-f5:
[NCC] Change service type HTTP to TCP for 443 passthrough traffic


subbuv26 (approved):
Looks Good

Issue #1391, opened 2020-07-07
Expose Kubernetes api services via F5 ingress crashes CIS


f5minions:
#### Setup Details
 CIS Version : 2.0      
 Build: f5networks/k8s-bigip-ctlr
 BIGIP Version: Big IP 14.1.2      
 AS3 Version: 3.20     
 Agent Mode: AS3       
 Orchestration: K8S     
 Orchestration Version: 1.17 and 1.18    
 Pool Mode: Cluster      
 Additional Setup details: <Platform/CNI Plugins/ cluster nodes/ etc>

#### Description  
when i configured ingress for kubernetes API server than CIS controller was stop working and become crashloopback status. I used below ingres YAML,

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    virtual-server.f5.com/health: |
      [
        {
          "path":     "kube-apiserver.skb-doj-dev01.btvpaas.com/",
          "send":     "HTTP GET /",
          "interval": 5,
          "timeout":  10
        }
      ]
    virtual-server.f5.com/ip: 175.113.150.30
    virtual-server.f5.com/partition: k8s-dev
  name: kube-apiserver
  namespace: default
spec:
  rules:
  - host: kube-apiserver.skb-doj-dev01.btvpaas.com
    http:
      paths:
      - backend:
          serviceName: kubernetes
          servicePort: 443
        path: /
```
 

And CIS controller logs is below,
Controller log-----------------------------------------------
 
[centos@doj-dw03 ~]$ sudo docker logs f1cff628b6f5
2020/07/06 12:28:27 [INFO] [INIT] Starting: Container Ingress Services - Version: 2.0.0, BuildInfo: cloud-user-efcab9e-20200522033624
2020/07/06 12:28:27 [INFO] ConfigWriter started: 0xc00030c480
2020/07/06 12:28:27 [INFO] Started config driver sub-process at pid: 16
2020/07/06 12:28:27 [INFO] [INIT] Creating Agent for as3
2020/07/06 12:28:27 [INFO] [AS3] Initializing AS3 Agent
2020/07/06 12:28:27 [INFO] [CCCL] Initializing CCCL Agent
2020/07/06 12:28:27 [INFO] [CORE] NodePoller (0xc00011eab0) registering new listener: 0x130d7c0
2020/07/06 12:28:27 [INFO] [CORE] NodePoller started: (0xc00011eab0)
[...]

PR #1575, opened 2020-11-17
Fix CIS crash when kubernetes api services are exposed

vklohiya:
Fix for [Github-1391](https://github.com/F5Networks/k8s-bigip-ctlr/issues/1391). Fix CIS crash when kubernetes api services are exposed


subbuv26 (approved):
Looks Good.

PR #1583, opened 2020-11-24
Added new cli parameters for external gtm

vklohiya:
Usage: 
--gtm-bigip-password string          Optional, password for the GMT Big-IP user account.
--gtm-bigip-url string               Optional, URL for the GTM Big-IP
--gtm-bigip-username string          Optional, user name for the GTM Big-IP user account.
--gtm-credentials-directory string   Optional, directory that contains the GTM BIG-IP username, password, and/or url files. To be used instead of username, password, and/or url arguments.


subbuv26 (approved):
Looks Good.

Issue #602, opened 2018-03-06
K8sBigipCtlr unit test suite intermittent timeout errors


ryan-talley:
#### Description
Unit tests for the k8s-bigip-ctlr started encountering intermittent timeout errors in the K8sBigipCtlr Suite after recent changes to this suite. The full error message is as follows:
```
[1520358326] K8sBigipCtlr Suite - 13 specs - 4 nodes ••••••••••••
	 -------------------------------------------------------------------
	|                                                                   |
	|  Ginkgo timed out waiting for all parallel nodes to report back!  |
	|                                                                   |
	 -------------------------------------------------------------------

k8s-bigip-ctlr timed out. path: ./cmd/k8s-bigip-ctlr
```
#### Controller Version
v1.5

sjberman:
Running this test locally (`make local-go-test`), I get this error:

```
• Failure [0.007 seconds]
Main Tests Main Tests [It] sets up the driver command
/home/ubuntu/go/src/github.com/F5Networks/k8s-bigip-ctlr/cmd/k8s-bigip-ctlr/main_test.go:131

  We should find the driver.
  Expected
      <*exec.Error | 0xc420128040>: {
          Name: "bigipconfigdriver.py",
          Err: {
              s: "executable file not found in $PATH",
          },
      }
  to be nil

  /home/ubuntu/go/src/github.com/F5Networks/k8s-bigip-ctlr/cmd/k8s-bigip-ctlr/main_test.go:136
```

dylan-way:
The above error found by running `make local-go-test` is now being tracked at #623 .

PR #1616, opened 2020-12-21
handle duplicate path for same hostname

lavanya-f5:
Handles duplicate paths with multiple virtual server custom resources pointing to same hostname and virtualaddress.

Fixes gitlab #602


srivastavnitin24 (approved):
looks fine

Issue #1602, opened 2020-12-07
CIS 2.2.1 now fails to complete AS3 declaration that worked with previous versions


mikeoleary:
#### Setup Details
 CIS Version : 2.2.1      
 Build: f5networks/k8s-bigip-ctlr:latest       
 BIGIP Version: Big IP 15.1.0.4     
 AS3 Version: 3.20
 Agent Mode: AS3
 Orchestration: K8S/OSCP      
 Orchestration Version:       
 Pool Mode: Cluster     
 Additional Setup details: EKS

#### Description  
With CIS 2.2.1, my AS3 declarations now get an error:
````
[/app1/helloworld/helloworld_pool/members/0: should have required property 'bigip'] message:declaration is invalid]
````

However if I take my AS3 declaration and POST it using Postman, I can create the application successfully. The declaration in the configmap appears to be valid.

This started happening with 2.2.1. I've tested the two previous versions (v2.2.0 and 2.1.1) and this was not a problem. The same AS3 declaration is fine.

#### Steps To Reproduce
1) I've used Terraform to deploy EKS and then install CIS and a demo app. This has worked for a few months but appears to have stopped working with the release of CIS 2.2.1

2) This is my ConfigMap in K8s
```yaml
ubuntu@ip-10-0-0-36:~/aws-multisite-vpn-f5/4-eks_build_out$ kubectl get cm -o yaml
apiVersion: v1
items:
- apiVersion: v1
  data:
    template: |2-
         {
              "class": "AS3",
              "declaration": {
                  "class": "ADC",
                  "schemaVersion": "3.10.0",
                  "id": "urn:uuid:33045210-3ab8-4636-9b2a-c98d22ab915d",
                  "label": "http",
                  "remark": "Simple HTTP application",
                  "app1": {
                      "class": "Tenant",
                      "helloworld": {
                          "class": "Application",
                          "template": "generic",
                          "helloworld_vs": {
                              "class": "Service_HTTP",
                              "virtualAddresses": [
                                  "192.168.2.160"
                              ],
                              "virtualPort": 80,
                              "remark":"ingress: better together demo app",
                              "pool": "helloworld_pool"
                          },
                          "helloworld_pool": {
                              "class": "Pool",
                              "monitors": [
                                  "http"
                              ],
                              "members": [
                                  {
                                      "servicePort": 80,
                                      "serverAddresses": []
                                  }
                              ]
                          }
                      }
                  }
              }
          }
[...]

PR #1620, opened 2021-01-04
Fix configmap serviceport value for nodeport mode #1602

sravyap135:
Problem : With multiport service support enabled from 2.2.1, while running CIS in nodeport mode, user has to provide service's nodeport value in the configmap servicePort.

Solution: Fix made to map service port to corresponding service's nodeport so that config maps are compatible for 2.2.0 and 2.2.1

Ref: 
https://github.com/F5Networks/k8s-bigip-ctlr/pull/1564 
https://github.com/F5Networks/k8s-bigip-ctlr/issues/1602


vklohiya (approved):
Looks fine to me.

PR #1622, opened 2021-01-05
Handle invalid configmap processing

lavanya-f5:
Problem: CIS deletes the previously existing configuration from tenant , if corresponding configmap is edited with invalid config.

Solution: Ignore the invalid config and dont cleanup configuration unless configmap is updated with valid config or deleted.


subbuv26 (approved):
Looks Good.

Issue #1671, opened 2021-01-28
TransportServer assigns wrong pool/service


chen23:
#### Setup Detail
 CIS Version : 2.2.3    
 Build: f5networks/k8s-bigip-ctlr:latest       
 BIGIP Version: Big IP 15.1.2
 AS3 Version: 3.24
 Agent Mode: AS3/CCCL       
 Orchestration: K8S/OSCP      
 Orchestration Version:       
 Pool Mode: Cluster/Nodeport      
 Additional Setup details: <Platform/CNI Plugins/ cluster nodes/ etc>

#### Description  
  
When deploying more than one transport server the controller will assign the wrong pool.  It creates the pool resources, but assigns the wrong pool.

#### Steps To Reproduce
1) Create two TS with the same virtualServerAddress and different ports, assign different services

#### Expected Result 
targets different backends

#### Actual Result  
it does not

#### Diagnostic Information
TS
```
$ kubectl get ts -n cluster2
NAME              VIRTUALSERVERADDRESS   VIRTUALSERVERPORT   POOL          POOLPORT   AGE
frontend1         10.1.10.83             8081                frontend1     80         5m46s
my-frontend-crd   10.1.10.83             8080                my-frontend   80         45h
```
GUI
![image](https://user-images.githubusercontent.com/63689/106118623-4e937100-6122-11eb-9d0c-df248e830f2d.png)

TMSH
```
ltm virtual Shared/crd_10_1_10_83_8080 {
    creation-time 2021-01-26:04:02:29
    description Shared
    destination 10.1.10.83:webcache
    ip-protocol tcp
    last-modified-time 2021-01-26:04:02:29
    mask 255.255.255.255
    partition cluster2
    persist {
        /Common/source_addr {
            default yes
        }
    }
    pool Shared/cluster2_my_frontend_80
    profiles {
        /Common/f5-tcp-progressive { }
    }
    serverssl-use-sni disabled
    source 0.0.0.0/0
    source-address-translation {
        type automap
    }
    translate-address enabled
    translate-port enabled
    vs-index 60
}
ltm virtual Shared/crd_10_1_10_83_8081 {
    creation-time 2021-01-28:01:26:28
    description Shared
    destination 10.1.10.83:tproxy
    ip-protocol tcp
    last-modified-time 2021-01-28:01:26:28
    mask 255.255.255.255
    partition cluster2
    persist {
        /Common/source_addr {
            default yes
        }
    }
[...]

PR #1674, opened 2021-01-29
Fixing the issue where TransportServer assigns wrong pool/service

vklohiya:
Problem: TransportServer assigns wrong pool/service when same IP address is used with two transport server as mentioned in [Github-1671](https://github.com/F5Networks/k8s-bigip-ctlr/issues/1671).
Solution: Added the logic for checking the combination of transport server's address and port are unique and the corresponding services are attached properly.


subbuv26 (approved):
Looks Good to me

subbuv26 (approved):
Looks good to me

PR #1675, opened 2021-02-01
Fix for mac address update in ARP

vklohiya:
Problem: CIS was unable to update the ARP entries on big ip when there is Mac address update instead of IP address.
Solution: Added a condition to remove the ipaddress field from payload when ever updating the ARP, as it's an immutable field.


subbuv26 (approved):
Looks Good

Issue #1641, opened 2021-01-18
Debug log sKey.ServiceName in syncVirtualServer


vincentmli:
#### Setup Details
 CIS Version : 2.2.1      
 Build: f5networks/k8s-bigip-ctlr:latest       
 BIGIP Version: Big IP x.x.x      
 AS3 Version: 3.x      
 Agent Mode: AS3/CCCL       
 Orchestration: K8S/OSCP      
 Orchestration Version:       
 Pool Mode: Cluster
 Additional Setup details: <Platform/CNI Plugins/ cluster nodes/ etc>

#### Description  

syncVirtualServer debug log should log sKey.ServiceName instead of sKey.Name, sKey.Name always result empty string in debug log

<pre>
diff --git a/pkg/appmanager/appManager.go b/pkg/appmanager/appManager.go
index 09ecadcb..cb1f0aff 100644
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -932,7 +932,7 @@ func (appMgr *Manager) syncVirtualServer(sKey serviceQueueKey) error {
        defer func() {
                endTime := time.Now()
                log.Debugf("[CORE] Finished syncing virtual servers %+v %+v (%v)",
-                       sKey.Name, sKey.Namespace, endTime.Sub(startTime))
+                       sKey.ServiceName, sKey.Namespace, endTime.Sub(startTime))
        }()
        // Get the informers for the namespace. This will tell us if we care about
        // this item.
  
</pre>

#### Steps To Reproduce

enable CIS debug 

#### Expected Result 

syncVirtualServer debug logs service name

#### Actual Result  

syncVirtualServer debug logs empty service name

#### Diagnostic Information

after fix:

2021/01/18 22:53:11 [DEBUG] [CORE] Finished syncing virtual servers nginxservice default (24.867µs)
2021/01/18 22:53:12 [DEBUG] [CORE] Finished syncing virtual servers f5-hello-world-end-to-end-ssl default (596.966236ms)
2021/01/18 22:53:12 [DEBUG] [CORE] Finished syncing virtual servers f5-hello-world-waf default (400.029624ms)
2021/01/18 22:53:12 [DEBUG] [CORE] Finished syncing virtual servers f5-hello-world-https default (399.890213ms)
2021/01/18 22:53:13 [DEBUG] [CORE] Finished syncing virtual servers f5-hello-world default (603.875016ms)

before fix:

[...]

PR #1677, opened 2021-02-01
Fix VS log message

sravyap135:
https://github.com/F5Networks/k8s-bigip-ctlr/issues/1641 

Problem : [DEBUG] [CORE] Finished syncing virtual servers default (24.867µs)

Solution:

[DEBUG] [CORE] Finished syncing virtual servers example_vs in namespace default (24.867µs)


agupta49 (approved):
Looks good to me

PR #1691, opened 2021-02-09
Support for caCertificate in TLS Edge termination route

agupta49:
Problem: OpenShift-Route Parameter "caCertificate" is ignored by CIS.

Solution:  CIS is able to support `caCertificate` param from OpenShift edge termination Routes using AS3 `ChainCA`.

Signed-off-by: : Amit Gupta amitg2812@gmail.com


subbuv26 (approved):
Looks Good.

Issue #1458, opened 2020-08-12
CISv2.1.0 does not delete LTM-Policy reset-rule when OpenShift-annotation for whitelist-source-range will be removed


joebride:
#### Setup Details
 CIS Version : 2.1.0   
 Build: f5networks/k8s-bigip-ctlr:latest       
 BIGIP Version: Big IP 15.1.0.3  
 AS3 Version: 3.20.0-build3 
 Agent Mode: AS3       
 Orchestration: K8S/OSCP      v1.17.1, OS v4.2
 Orchestration Version:       
 Pool Mode: Cluster 
 Additional Setup details: OpenShift-Routes using virtual-server.f5.com/whitelist-source-range: 

#### Description  
When I remove the annotation "virtual-server.f5.com/whitelist-source-range:" from my OpenShift-Route, the CIS does NOT update the matching rule in the BIG-IPs LTM-Policy (openshift_secure_routes, openshift_insecure_routes) correctly.
It does NOT delete the following reset-rule (example):
osr_f5test_httpd_b_example_reset {
            actions {
                0 {
                    shutdown
                    connection
                }
                1 {
                    asm
                    disable
                }
            }
            conditions {
                0 {
                    http-header
                    name host
                    values { <hostname> }
                }
            }
            ordinal 4
        }

#### Steps To Reproduce
1) add annotation "virtual-server.f5.com/whitelist-source-range: ..." to the relevant OpenShift-Route
2) remove annotation "virtual-server.f5.com/whitelist-source-range: ..." to the relevant OpenShift-Route
3) check LTM-policy on BIG-IP

#### Expected Result 
remove reset-rule in LTM-Policy (openshift_secure_routes, openshift_insecure_routes)

#### Actual Result  
Does NOT remove reset-rule in LTM-Policy (openshift_secure_routes, openshift_insecure_routes)

#### Diagnostic Information

Excerpt CIS-config
      image: 'registry.connect.redhat.com/f5networks/cntr-ingress-svcs:2.1.0-rhel7'
      args:
        - '--bigip-username=$(BIGIP_USERNAME)'
        - '--bigip-password=$(BIGIP_PASSWORD)'
        - '--bigip-url=A.B.C.D'
        - '--bigip-partition=XXX'
        - '--pool-member-type=cluster'
        - '--openshift-sdn-name=/Common/RD_000_TUN_VXL_0001'
        - '--log-level=INFO'
        - '--log-as3-response=true'
[...]

PR #1706, opened 2021-02-16
Remove reset rule when whitelist range is updated

sravyap135:
**Description**:  
https://github.com/F5Networks/k8s-bigip-ctlr/issues/1458 
Reset rule exists in BIGIP even after deleting a whitelist/source-range annotation

**Changes Proposed in PR**:
Delete reset matching rule from policy and update config

**Fixes**: #1458 

## General Checklist
- [x] updated release notes for bug fix

### DIFF
--- a/pkg/appmanager/appManager.go
+++ b/pkg/appmanager/appManager.go
@@ -1384,9 +1384,6 @@ func (appMgr *Manager) syncIngresses(
 						}
 					}
 				}
-				if dep.Kind == WhitelistDep {
-					rsCfg.DeleteWhitelistCondition(dep.Name)
-				}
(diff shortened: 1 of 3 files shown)


pkg/appmanager/appManager.go
```diff
@@ -1384,9 +1384,6 @@ func (appMgr *Manager) syncIngresses(
 						}
 					}
 				}
-				if dep.Kind == WhitelistDep {
```
sravyap135, line 1387: Ingress resource doesn't have any reset rule associated like for routes

agupta49 (approved):

PR #1708, opened 2021-02-17
Fixing attribution generator issue

vklohiya:
**Description**:  
Currently "make prod" fails with attribution generation error.

**Changes Proposed in PR**:
Introduced the conditional based attribution generation. To generate the all_attribution.txt, you can run command "make prod-license"


subbuv26 (approved):
Looks Good

PR #1709, opened 2021-02-17
Allow user to leverage IPAM

agupta49:
**Description**:  Allow user to leverage IPAM Controller or specify a VirtualServer Address.

**Changes Proposed in PR**: 
We need to provide more flexibility
        if a VirtualServer Address is specified in the resource, don't leverage the IPAM Controller even if a cidr parameter is specified
        if No VirtualServer Address is specified in the resource and a cidr parameter is specified, leverage the IPAM Controller

## General Checklist


vklohiya (approved):
Looks Fine

PR #1766, opened 2021-03-25
Fix for implicit health monitor with ingressLink resource

lavanya-f5:
**Description**:  Support for implicit health monitor with ingressLink resource in nodeport mode


vklohiya (approved):
Looks Fine.

PR #1770, opened 2021-03-30
Upgrading the vendor packages to support k8s 1.20

vklohiya:
Description: Upgrading the vendor packages to support k8s 1.20

Changes Proposed in PR: We are upgrading the kube client to v0.20.4 and other related dependencies.


vklohiya:
Closing

iam-veeramalla:
@vklohiya Should you instead consider deleting the vendor folder itself? The reason why we had it there was while upgrading from GO deps to GO mods we want to keep vendor for atleast couple of Go releases. There are many good reasons to get rid of vendor in GitHub. 

Vendor folder is simply like maintaining a version control system in another version control system(GitHub).
They make PR's look huge and waste a lot of space(Microsoft to worry :p)
No real advantage.

While we might argue on stuff like:
What if GitHub is Down?
What if the version mentioned in go.mod is not available(probably deleted from GitHub). 

But practically 1 is very less likely to happen and 2 is arguable. 

Note: Most of the popular projects do not push vendor to their source code repository.

PR #1771, opened 2021-03-30
Upgrading the vendor packages to support k8s 1.20

vklohiya:
Description: Upgrading the vendor packages to support k8s 1.20

Changes Proposed in PR: We are upgrading the kube client to v0.20.4 and other related dependencies.


subbuv26 (approved):
Looks Good

Issue #1737, opened 2021-03-05
Inconsistent ordering of policy rules when adding an Ingress path


conkiztador:
#### Setup Details
 CIS Version : 2.X      
 Build: f5networks/k8s-bigip-ctlr:latest       
 BIGIP Version: Big IP x.x.x      
 AS3 Version: 3.x      
 Agent Mode: AS3       
 Orchestration: K8S      
 Orchestration Version:       
 Pool Mode: Nodeport      

#### Description  

According to the Ingress docs (https://kubernetes.io/docs/concepts/services-networking/ingress/#multiple-matches):

> In some cases, multiple paths within an Ingress will match a request. In those cases precedence will be given first to the longest matching path. If two paths are still equally matched, precedence will be given to paths with an exact path type over prefix path type.

The F5 controller uses the "first matching" strategy for the vserver policy, so the ordering of rules matter. When an Ingress is first created the rules seem to be ordered appropriately, but if more services are added to the Ingress later, they are just added to the end of the policy.

This is a problem particularly for things like `cert-manager`, which adds a temporary path like `/.well-known/acme-challenge/xxxx` to an Ingress for http01 challenge verification.

#### Steps To Reproduce

Example setup:
```apiVersion: v1
kind: Pod
metadata:
  labels:
    app: nginx1
  name: nginx1
spec:
  containers:
  - image: nginx
    name: nginx
    args:
      - "echo one > /usr/share/nginx/html/index.html; exec nginx -g 'daemon off;'"
    command:
      - bash
      - -c
    ports:
    - containerPort: 80
      protocol: TCP
---
apiVersion: v1
kind: Pod
metadata:
  labels:
    app: nginx2
  name: nginx2
spec:
  containers:
  - image: nginx
    name: nginx
    args:
      - "mkdir /usr/share/nginx/html/two; echo two > /usr/share/nginx/html/two/index.html; exec nginx -g 'daemon off;'"
    command:
      - bash
      - -c
    ports:
    - containerPort: 80
      protocol: TCP
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: nginx1
  name: nginx1
spec:
  ports:
  - name: http
    nodePort: 31234
    port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: nginx1
  type: NodePort
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: nginx2
  name: nginx2
spec:
  ports:
  - name: http
[...]

PR #1800, opened 2021-04-22
Fix for Inconsistent ordering of policy rules when adding an Ingress path

vklohiya:

**Description**: Fix for Inconsistent ordering of policy rules when adding an Ingress path

**Changes Proposed in PR**: Sorted the order of rules before pushing to bigip

**Fixes**: #1737

## General Checklist


subbuv26 (approved):
Looks Good

PR #1821, opened 2021-05-11
Fix Processing endpoint with node names when CIS in Cluster mode

vklohiya:
**Description**: Fix Processing endpoint with node names when CIS in Cluster mode

**Changes Proposed in PR**: Removed the node check while adding the endpoints.

## General Checklist


subbuv26 (approved):
Looks Good.

PR #1860, opened 2021-06-11
Fix for CIS adds extra slash '/' when set rewrite-target-url annotation as '/'

vklohiya:
Fix for CIS adds extra slash '/' when set rewrite-target-url annotation as '/'.

## General Checklist


subbuv26 (approved):
Looks Good.

agupta49 (approved):

Issue #1921, opened 2021-07-27
Plain text login and password in process status on node that is running controller


tomaszmzyk:
#### Setup Details
 CIS Version : 2.2.1, 2.5.0   
 Build: f5networks/k8s-bigip-ctlr:latest       
 BIGIP Version: Big IP 15.1.3      
 AS3 Version: 3.18      
 Agent Mode: AS3/CCCL       
 Orchestration: OSCP      
 Orchestration Version: 3.11, 4.6
 Pool Mode: Cluster    
 Additional Setup details: OpenshiftSDN

#### Description  
 When you deploy a controller on Openshift according to documentation:
https://clouddocs.f5.com/containers/latest/userguide/openshift/

Pod running:
NAME                                 READY   STATUS    RESTARTS   AGE
ocp-k8s-bigip-ctrl-95864f589-jms4d   1/1     Running   0          4d22h

But when you login to node, that pod is running in process output, for example by command :
ps aux | grep k8s you see plain texr credential for user from F%, which one have a Administrator role.

for example
core      228154  0.1  0.1 728964 57524 ?        Ssl  Jul22  13:23 /app/bin/k8s-bigip-ctlr.real --bigip-username=<User> --bigip-password=<PlainTextPassword> --bigip-url=<IP> --bigip-partition=T1_OCP4 --pool-member-type=cluster --openshift-sdn-name=/part/vxlan --log-level=DEBUG --log-as3-response=true --insecure=true --verify-interval=120

#### Steps To Reproduce
1) Create deployment for controller
2) Run deployment
3) Check where the pod is running and watch ps command output for k8s controller

#### Expected Result 

Login, Password and even IP should be encrypted

#### Actual Result  
All sensitive data are in plain text

#### Diagnostic Information
```buildoutcfg
<Configuration files, error messages, logs>
Note: Sanitize the data. For example, be mindful of IPs, ports, application names and URLs
Note: The following F5 article outlines the information required when opening an issue.
https://support.f5.com/csp/article/K60974137
```

vklohiya:
[...]

PR #1927, opened 2021-07-29
Fixing the issue with credential-directory parameter

vklohiya:
**Description**:  If there is newline character at end of file due to which CIS fails to map the username, password and url. We need to remove the newline character while parsing file.

**Changes Proposed in PR**: Removing the newline characters from file while mapping the credentials. Customer can use --credentials-directory parameter instead of these three parameters (--bigip-username, --bigip-password, --bigip-url ). For that they need to create three files named username, password and url inside a directory and then create a Kubernetes secrets using this directory. Please see the CIS deploy file for [example](https://raw.githubusercontent.com/F5Networks/k8s-bigip-ctlr/master/docs/config_examples/example-bigip-credentials-directory.yaml)

**Fixes**: #1921

## General Checklist

### DIFF
--- a/cmd/k8s-bigip-ctlr/main.go
+++ b/cmd/k8s-bigip-ctlr/main.go
@@ -552,7 +552,7 @@ func getCredentials() error {
 					return fmt.Errorf(fmt.Sprintf("BIG-IP %s not specified", fieldType))
 				}
 			} else {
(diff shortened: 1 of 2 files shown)


subbuv26 (approved):
LGTM

PR #1951, opened 2021-08-16
Using fmt.Sprintf alternatives to improve performance

nandakishorepeddi:
**Description**:  _Some places in the code are using Sprintf in combination with log library which can be avoided 

`fmt.Sprintf` is great but it needs to marshal its variable arguments into a slice of interface{}. The compiler has improved, but this could cause unnecessary allocations. There are simpler functions for the same task that don’t need variable arguments or interface arguments._

**Changes Proposed in PR**:

## General Checklist

## CRD Checklist


agupta49 (approved):
LGTM

nandakishorepeddi:
azp /run

PR #1972, opened 2021-08-24
Process only associated EDNS when new VS gets created

sravyap135:
**Description**:  
When a new virtual server is created, then associated EDNS resources need to be processed. But, currently all EDNS resources are getting processed. 

**Changes Proposed in PR**: Using the hosts that the VS handles, EDNS resources can be filtered, and only the filtered ones are processed.

**Fixes**: resolves CONTCNTR-2895

## General Checklist
- [ ] Added examples for new feature - NA
- [ ] Updated the documentation - NA

## CRD Checklist
- [ ] Updated required CR schema - NA


agupta49:
Please create two separate PR's for make file changes and edns changes.

PR #2089, opened 2021-11-12
LTM resource name changes for multihost support

sravyap135:
**Description**:  
LTM resource names are framed based on hostname in certain places. Now this need to be based on context, in case of single host, hostname can be used, but in case of Multi-host hostGroup can be used.

**Changes Proposed in PR**:
- Helper functions identify the context and frame LTM resource name accordingly.


agupta49 (approved):
LGTM

PR #2103, opened 2021-11-18
Fix firewall and log profile support for transport server

sravyap135:
**Description**:  
Policy CRD Firewall and log profile assignment for transport server was missed while framing AS3 declaration

**Changes Proposed in PR**:
- common AS3 supported options in VS and TS are now processed in `processCommonDecl` function

**Fixes**: internal CONTCNTR-3022

## General Checklist


vklohiya (approved):
LGTM

PR #2145, opened 2021-12-16
Fix VS processing with multihost mode

sravyap135:
**Description**:  Fix VS processing with multihost mode

**Changes Proposed in PR**:
- return and log error only when it is non-nil

**Fixes**: internal 792


vklohiya (approved):
LGTM

agupta49 (approved):

PR #2148, opened 2021-12-16
Fix for  Allow type LoadBalancer with different TargetPort and Port value in nodeport mode

vklohiya:
Fix for  Allow type LoadBalancer with different TargetPort and Port value in nodeport mode

## General Checklist


nandakishorepeddi (approved):
LGTM

PR #2149, opened 2021-12-16
Modify serviceTypeLB naming convention to support ipv6 addresses

sravyap135:
**Description**:  Modify serviceTypeLB naming convention to support ipv6 addresses

**Changes Proposed in PR**:
- Modify serviceTypeLB naming convention to support ipv6 addresses


vklohiya (approved):
LGTM

PR #2150, opened 2021-12-16
Handle VS,TS status update when they are not found

sravyap135:
**Description**: Handle VS,TS status update when they are not found

**Changes Proposed in PR**:
- break when there are error 

**Fixes**: internal 794


nandakishorepeddi (approved):
LGTM

PR #2155, opened 2021-12-21
Fix teem CRD initialisation

sravyap135:
**Description**:  Sometimes, even before teem data is initialised, CRManager is trying to modify it. This is resulting in a crash. This will not be an issue in appmanager as informers are setup after teemdata initialisation. 

**Changes Proposed in PR**:
- Initialise Teem data immediately after starting CR informers, even before get BIGIP registration key request.
- Also, updating latest FIC go mod

**Fixes**: internal 791, 800


subbuv26 (approved):
LGTM

PR #2156, opened 2021-12-21
Cleanup code to delete default _as3 partition CCCL and as3

sravyap135:
**Description**:  Cleanup code to remove default  _as3 partition CCCL and as3

**Changes Proposed in PR**:
- Remove code which removed _AS3 partition in CCCL mode and as3 mode


subbuv26 (approved):
LGTM

Issue #2077, opened 2021-11-09
CIS deletes all existing ARP on restart and recreates it, which affects traffic


kkfinkkfin:
#### Setup Details
CIS Version : 2.6.1
Build: f5networks/k8s-bigip-ctlr:latest
BIGIP Version: BIG-IP 15.1.4 Build 0.0.47 Final
AS3 Version: none
Agent Mode: CCCL
Orchestration: K8S
Orchestration Version: kubernetes v1.21.5
Pool Mode: Cluster
Additional Setup details: <Platform/CNI Plugins/ cluster nodes/ etc>

Platform : CentOS Linux release 8.4.2105
Kernel: 4.18.0-305.19.1.el8_4.x86_64
CNI Plugins: flannel

#### Steps To Reproduce
1) restart CIS ，this is the yaml 
```CIS yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: cc-k8s-to-bigip1
  namespace: kube-system
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: k8s-bigip-ctlr
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      labels:
        app: k8s-bigip-ctlr
    spec:
      containers:
      - args:
        - --bigip-username=admin
        - --bigip-password=admin.F5demo.com
        - --bigip-url=10.1.20.232
        - --bigip-partition=p1
        - --pool-member-type=cluster
        - --flannel-name=/Common/flannel_vxlan
        - --insecure=true
        - --log-level=INFO
        - --agent=cccl
        - --verify-interval=5
        - --node-poll-interval=5
        - --http-listen-address=0.0.0.0:9113
        - --manage-ingress=true
        command:
        - /app/bin/k8s-bigip-ctlr
        image: f5networks/k8s-bigip-ctlr:2.6.1
        imagePullPolicy: IfNotPresent
        name: k8s-bigip-ctlr
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      serviceAccount: bigip-ctlr
      serviceAccountName: bigip-ctlr
      terminationGracePeriodSeconds: 30
```  
2) see the CIS log

#### Expected Result 
The ARP record for an existing pod is not deleted.

#### Actual Result  
[...]

PR #2173, opened 2022-01-08
Fix the issue of config rebuilding when controller is restarted.

zongzw:
**Description**:  

Fix the issue of config rebuilding when controller is restarted.

**Changes Proposed in PR**:

Read the existing config.json(if exists) and use the content as the original input.
Besides, if user want to make sure the existing resources on BIG-IP not be removed, they need to mount /tmp/k8s-bigip-ctlr folder as the volume.

**Fixes**: resolves #_2077_

## General Checklist
- [ ] Added examples for new feature
- [ ] Updated the documentation
- [ ] Smoke testing completed

## CRD Checklist

### DIFF
--- a/pkg/writer/configWriter.go
+++ b/pkg/writer/configWriter.go
@@ -19,7 +19,6 @@ package writer
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"syscall"
@@ -62,9 +61,14 @@ type doneCall func(chan<- struct{})
 type errCall func(chan<- error, error)
 
 func NewConfigWriter() (Writer, error) {
-	dir, err := ioutil.TempDir("", "k8s-bigip-ctlr.config")
+	dir := filepath.Join(os.TempDir(), "k8s-bigip-ctlr.config")
+	_, err := os.Stat(dir)
 	if nil != err {
-		return nil, fmt.Errorf("could not create unique config directory: %v", err)
+		if os.IsNotExist(err) {
+			os.Mkdir(dir, os.ModePerm)
+		} else {
+			return nil, fmt.Errorf("could not create unique config directory: %v", err)
+		}
 	}
 
 	tmpfn := filepath.Join(dir, "config.json")
@@ -73,8 +77,19 @@ func NewConfigWriter() (Writer, error) {
 		configFile: tmpfn,
 		stopCh:     make(chan struct{}),
 		dataCh:     make(chan configSection),
-		sectionMap: make(map[string]interface{}),
+		sectionMap: nil,
 	}
+
+	conf := make(map[string]interface{})
+	b, err := cw.lockAndRead()
+	if nil != err {
+		log.Warningf("[CCCL] Failed to read from %s: %v", cw.configFile, err)
+	}
+
+	if err = json.Unmarshal(b, &conf); nil != err {
+		log.Warningf("[CCCL] Invalid content from %s", cw.configFile)
+	}
+	cw.sectionMap = conf
 
 	go cw.waitData()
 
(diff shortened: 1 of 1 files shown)


zongzw:
> @zongzw - regarding the PR solution,
> 
> > Read the existing config.json(if exists) and use the content as the original input.
> > Besides, if user want to make sure the existing resources on BIG-IP not be removed, they need to mount /tmp/k8s-bigip-ctlr folder as the volume.
> 
> IMO, While CIS is restarted and if there are changes to CIS monitored resources in the clusters, this fix will not be helpful.

I tested with the following steps, but didn't reproduce your assertion. Appreciate if you can show the details.
Could you please provide an example or test procedures to show 'will not be helpful'?

1) Use `watch -n 1 "curl -k -u admin:P@ssw0rd123 https://<BIG-IP MGMT IP>:8443/mgmt/tm/net/arp/ 2>/dev/null | python -m json.tool | grep ipAddress "` to monitor ARP list on BIG-IP.
2) Start CIS and create deployment(replica=1)/service/ingress via kubectl, check resource creation on BIG-IP, thus, assert CIS works well.
3) Stop CIS and scale out deployment replicas from 1 to 10.
4) Start CIS and observe from 1)'s ARPs changes from 1 to 10, not  1 -> 0 -> 10, actually 1 -> x1 -> x2 -> 10.
5) Stop CIS and scale in deployment replicas from 10 to 1.
6) Start CIS and observe from 1)'s ARPs changes from 10 -> x3 -> 1 not 10 -> 0 -> 1.
7) Keep CIS running and scale out deployment replicas from 1 -> 20, observing that CIS works well.

trinaths:
@zongzw - here, 

> they need to mount /tmp/k8s-bigip-ctlr folder as the volume.

This changes UX. This change contradicts many users using CIS seamlessly. 

Also, is this fix helpful only for iapps ?

zongzw:
> @zongzw - here,
> 
> > they need to mount /tmp/k8s-bigip-ctlr folder as the volume.
> 
> This changes UX. This change contradicts many users using CIS seamlessly.
> 
> Also, is this fix helpful only for iapps ?

Thanks for the comments. 
First, to the question, I verified the change for both as3 and iapp. They work OK.
However, I'm still a little confusing about your former concern on 'changes to CIS monitored resources in the clusters' while restart.

To the UX change, your concern is reasonable. When I pretended to code on it, I also considered it. 
However, in my opinion to the working frame, because CIS is using an external file as a state storage, the CIS runtime can not be regard as stateless actually(at least cccl side). Resuming from the point of restart instead of throwing away the existing config.json is more likely 'it should be'.
The state is a just piece of info that it can be stored in mounted volume or stored as configmap etc..Saying alot is useless, your acceptance to the config.json being state/useful is the most of all.

zongzw:
> @zongzw
> 
> > 'changes to CIS monitored resources in the clusters' while restart.
> 
> I mean, when CIS is down, and monitored resources like, configMaps, Ingress/Routes and CRDs are changing. When CIS back, considering config.json and not sending the cluster state to BIGIP is a concern here. Hope this is clear now.

The description is clear. You may have notice my description about 'changes to CIS monitored resources in the clusters' verification with deployment, right? From 1 to 7.

I have verified it with deployment scaling out/in. That trigger arp list changes so that we can observe the cccl behaviors while restarting CIS. 

Does it answer the question about 'changes to CIS monitored resources in the clusters'? 
What do you think about this 7 steps for 'changes to CIS monitored resources in the clusters'?

Since to 'configMaps, Ingress/Routes and CRDs', if CIS cannot sync/handle them onto BIG-IP when CIS is started. It would be the critical bug. To be honest, I've also verifyed configmap in CCCL mode. It works too 'while restart'. In AS3 mode, it has no relationship with config.json.

> 
> From an UX perspective, we cannot have user mount a volume. We need to be careful with external storage. This would be a tricky option to end users. Moreover, having sensitive data in external storage can be a security concern.

If you agree that CCCL is a stateful runtime, there must be a place to store the piece of info. However, external storage is just one option.

> 
> IMO, recommend improve this SR for further review or create a github issue or SR so that we will review the issue and fix the same. Please share detailed issue description and proposed solutions so that we can better understand the fix and help you.

My PR is aimed to solve the problem as #2077. The PR is not so complicated to understand, I think.

trinaths:
@zongzw 

> I've also verifyed configmap in CCCL mode. It works too 'while restart'. In AS3 mode, it has no relationship with config.json.

Can you please share your findings with changes to Ingress and routes in CCCL mode. With this change other resources are working well. 

> My PR is aimed to solve the problem as #2077. The PR is not so complicated to understand, I think.

volume mount and decision based on config.json are key points from the PR which need further testing to see how this effects UX.  Does this PR for #2077 https://github.com/F5Networks/k8s-bigip-ctlr/pull/2177 resolve the issue ? Please share your findings.

zongzw:
'With this change other resources are working well', Just wondering if the change #2177 has the fully corresponding tests to 'monitored resources like, configMaps, Ingress/Routes and CRDs'?

trinaths:
@zongzw - yes, we do complete regression for any PR.

zongzw:
> @zongzw - yes, we do complete regression for any PR.

Appreciate show me the test process and test result. I'll follow the tests to this PR.

trinaths:
Created CONTCNTR-3137 for internally tracking to validate this issue based on fix to CONTCNTR-3081.

sravyap135:
@zongzw Existing resources will no longer be removed from BIG-IP. Can you please try CIS 2.8.0 release and share your feedback.

zongzw:
Thanks. Closed it.

PR #2174, opened 2022-01-10
Support multihost VS policy rules for same path and service backend combination

sravyap135:
**Description**:  Support multihost VS policy rules for same path and service backend combination. 

- CIS with NGINX KIC for multiple hosts with default path /  

**Changes Proposed in PR**:
- append hostnames with an existing hostgroup rule. 

**Fixes**: internal CONTCNTR-3060

## General Checklist
- [x] Smoke testing completed


subbuv26 (approved):
LGTM

PR #2182, opened 2022-01-13
Added support for SCTP protocol in Services of type LoadBalancer

vklohiya:

**Description**:  Added support for SCTP protocol in Services of type LoadBalancer

**Changes Proposed in PR**: Added support for SCTP protocol in Services of type LoadBalancer

## General Checklist
- [x] Smoke testing completed

### DIFF
--- a/pkg/controller/backend.go
+++ b/pkg/controller/backend.go
@@ -981,6 +981,8 @@ func createTransportServiceDecl(cfg *ResourceConfig, sharedApp as3Application) {
 	if cfg.Virtual.Mode == "standard" {
 		if cfg.Virtual.IpProtocol == "udp" {
 			svc.Class = "Service_UDP"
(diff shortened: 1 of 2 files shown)


pkg/controller/backend.go
```diff
@@ -981,15 +981,19 @@ func createTransportServiceDecl(cfg *ResourceConfig, sharedApp as3Application) {
 	if cfg.Virtual.Mode == "standard" {
 		if cfg.Virtual.IpProtocol == "udp" {
 			svc.Class = "Service_UDP"
-		} else {
+		} else if cfg.Virtual.IpProtocol == "tcp" {
```
sravyap135, line 984: This would be a breaking change for existing users. Should we have tcp as default IP protocol ?   @trinaths

sravyap135 (approved):

PR #2186, opened 2022-01-17
Fix VS with same virtual address but different hosts

sravyap135:
Signed-off-by: Sravya Pondugula <sravyap135@gmail.com>

**Description**:  

When root domain and wildcard domain refer to same VSaddress, CIS is only pushing the latest updated resource config, it is overwriting the policies/rules. 

**Changes Proposed in PR**:

- Added an error log stating config fault and it stops further processing. 
- Fixed existing unit testcases and added new one. 

**Fixes**: resolves internal 805


subbuv26 (approved):
LGTM

PR #2190, opened 2022-01-19
Revert "Fix VS with same virtual address but different hosts"

sravyap135:
Reverts F5Networks/k8s-bigip-ctlr#2186

Reverting this PR to avoid hybrid - IPAM and static IP address combination -  deployment conflicts


lavanya-f5 (approved):
LGTM

PR #2217, opened 2022-02-03
Added fix for concurrent read/write of iRuleMap

vklohiya:

**Description**: Added fix for cis crash with routes
**Changes Proposed in PR**: Using the mutex lock for iRuleMap to avoid the concurrent read/write

## General Checklist


sravyap135 (approved):
LGTM

PR #2246, opened 2022-02-21
Updated the fix for processing oldest route when same host and path in routes

vklohiya:

**Description**:  Updated the fix for processing oldest route when same host and path in routes

## General Checklist
- [x] Smoke testing completed


nandakishorepeddi (approved):

subbuv26 (approved):
LGTM

PR #2264, opened 2022-03-07
Native controller insecure routes

sravyap135:
**Description**:  Native controller insecure routes

**Changes Proposed in PR**:
- insecure routes in native controller 
- made few methods generic to handle VS and routes


subbuv26 (approved):
Lgtm

PR #2270, opened 2022-03-09
Added support for MVIP with WAF

arzzon:
**Description**: Added support for MVIP with WAF

**Changes Proposed in PR**: Added support for MVIP with WAF


subbuv26 (approved):
LGTM

PR #2276, opened 2022-03-10
Fix for CIS not retrying when AS3 response code is 422

Vidhi-Pat:
**Description**: Fix for CIS is not retrying when AS3 response code is 422

**Changes Proposed in PR**: Fix for CIS is not retrying when AS3 response code is 422


sravyap135 (approved):
LGTM

PR #2322, opened 2022-03-24
Performance related: Use faster List from client-go/listers/core/v1

Niklaus-xie:
**Description**: Because this kind of list was called many times, if using
the old way, it takes more than 1 min to deploy the
configurations onto the bigip when the resource number
gets > 100. Using the new way minimize the time to about
10 seconds.  I used 'go mod vendor' to make vendored copy
of dependencies, otherwise, only service.go and
expansion_generated.go are needed. If we don't need vendor
anymore, we can remove that.

**Changes Proposed in PR**: For improve performance.

**Fixes**: Related to performance.

- [x] Smoke testing completed


trinaths:
Created [CONTCNTR-3242] for internal tracking.

trinaths:
Created [CONTCNTR-3287] for internal tracking.

lavanya-f5 (approved):
LGTM

Niklaus-xie:
resolved conflicts and rebased. Many thanks to all.   Thanks for the review @lavanya-f5

PR #2342, opened 2022-04-06
ConfigMap backward compatibility fix

### LINKED ISSUE #2255: Backwards incompatible in multiple port service
At the end of #1464,  Backwards incompatible issue has be raised, I would like to create another issue for tracking this issue.

The following are copied from #1340 

**Backwards incompatible issue** - In Previous version, the servicePort is optional, but in later version servicePort is mandatory, which cause upgrade issue, we spent logs of time add hub mode back for smooth upgrade, but now blocked on servicePort.

Could you help to re-investigate, at least made the servicePort is optional? made the velidation logic like：if the servicePort match the targetPort then use the new added logic, or else, keep the previous logic?

vklohiya:

**Description**:  Backwards incompatible in multiple port service

**Changes Proposed in PR**: If the service port does not match with the configMap port we are adding the first port of the service in the endpoint, this will make sure that old configMaps keeps working like earlier. 

**Fixes**: resolves #2255 

## General Checklist
- [x] Smoke testing completed

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -7,6 +7,7 @@ Enhancements
 ````````````
 * :issues:`2030` Fix for changes on Ingress Resource Service Port not reflected on BIGIP
 * :issues:`2205` Fix for deleting multiple EDNS resources
+* :issues:`2255` Fix for ConfigMap backward compatibility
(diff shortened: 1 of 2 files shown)


lavanya-f5 (approved):
Looks goog

PR #2352, opened 2022-04-08
Revert "Update go.mod to v2"

sravyap135:
Reverts F5Networks/k8s-bigip-ctlr#2351

reverting as go get still fails to fetch v2 from remote


nandakishorepeddi (approved):
LGTM

PR #2358, opened 2022-04-13
Fix issue of deleting partition when bad as3 schema.

zongzw:
**Description**:  

CIS should not delete existing config when configmap in bad format 

**Changes Proposed in PR**:

The issue happens because:
The coming configmap json(containing partitions to create) is fully parsed because of bad schema format.
Thus, by comparing existing configmap in CIS, the fails-to-parsed partition is considered as `deleteTenants`.

The fix can be: recording the parsing result, more concisely tell later logic to distinguish it's invalid or real-missing.

**Fixes**: resolves SR  1-8181523521 / connell-1 

## General Checklist
- [x] Updated Added functionality/ bug fix in release notes - TBD in release notes 
- [x] Smoke testing completed

### DIFF
--- a/pkg/agent/as3/as3Utils.go
+++ b/pkg/agent/as3/as3Utils.go
@@ -214,7 +214,7 @@ func getTenants(decl as3Declaration, includeEmptyTenant bool) []string {
 func getAS3ObjectFromTemplate(
 	template as3Template,
 ) (as3Object, bool) {
-	var tmpl interface{}
+	var tmpl map[string]interface{}
 	err := json.Unmarshal([]byte(template), &tmpl)
 	if err != nil {
 		log.Errorf("[AS3] JSON unmarshal failed: %v  %v", err, template)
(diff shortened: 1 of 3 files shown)


pkg/agent/as3/as3Utils.go
```diff
@@ -312,6 +320,13 @@ func getClass(obj interface{}) string {
 	return cl.(string)
 }
 
+func assertToBe(kind string, obj interface{}) bool {
+	if obj == nil {
+		return false
+	}
+	return (reflect.TypeOf(obj).Kind().String() == kind)
```
sravyap135, line 327: reflect package adds additional overhead. Can you revert these changes from this PR.

zongzw: The impact is 'ns'(nano seconds) level, but the deployment of declaration is 's' (seconds, in fact tens of seconds) level. They are totally not in the same page. 
A bit more, assertTobe(map)  can justify not only `map[string]interface{}` but also `map[string]string` or map[string]xxxx. It may have more usage abilities.

sravyap135: Benchmarking results - 

BenchmarkInterfaceTest-12 1000000000 **0.2527** ns/op 0 B/op 0 allocs/op

BenchmarkAssertToBe-12 482082412 **2.494** ns/op 0 B/op 0 allocs/op

-  reflect adds an additional overhead by a factor of 10X

sravyap135:
Could you please update the PR with latest master so that I can initiate smoke test on this and we can target to merge it in 2.9.1

sravyap135:
@zongzw I ran the smoke tests. Can you please rebase the PR, update release notes. I can trigger azure pipeline.

sravyap135 (approved):

PR #2368, opened 2022-04-20
LoadBalancingMethod support for VS,TS CRs

### LINKED ISSUE #2224: Selecting Load Balancing method on VS CRD
#### Title
The user should be able to define the LB method for the VirtualServer CRD similar to what we can do with Ingress

#### Description
The user should be able to define the LB method for the VirtualServer CRD similar to what we can do with Ingress

#### Actual Problem
There are many cases that Round Robin is not the most effective algorithm for LB. 

#### Solution Proposed
On the Pool spec we need to give the option to define the LB method. This needs to take place in the pool as you might want to have different LB methods per pool (as it is done on the BIGIP)

image

nandakishorepeddi:
**Description**:  LoadBalancingMethod support for VirtualServer and TransportServer CRs.

**Changes Proposed in PR**:

**Fixes**: resolves #2224 

## General Checklist
- [x] Smoke testing completed

## CRD Checklist

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -6,8 +6,13 @@ Next Release
 Added Functionality
 ```````````````````
 
+**What’s new:**
+    * CRD:
+        * LoadBalancingMethod support for VirtualServer and TransportServer CRs. See `Examples <https://github.com/F5Networks/k8s-bigip-ctlr/tree/master/docs/config_examples/customResource/>`_
+
 Bug Fixes
 ````````````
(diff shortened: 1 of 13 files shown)


docs/RELEASE-NOTES.rst
```diff
@@ -1,13 +1,18 @@
 Release Notes for Container Ingress Services for Kubernetes & OpenShift
 =======================================================================
 
-Next Release
+2.9.0
```
vklohiya, line 4: We usually put version 2.9.0 after the release.

vklohiya (approved):
LGTM

PR #2378, opened 2022-04-27
Added support for passthrough traffic termination for routes 

vklohiya:

**Description**:  Added support for passthrough traffic termination for routes 

**Changes Proposed in PR**: All the passthrough traffic will be handled via iRule only for both virtual server and route resources. Removed the policy rules implementation for virtual server's passthrough traffic.

## General Checklist
- [ ] Smoke testing completed


subbuv26 (approved):
LGTM

PR #2380, opened 2022-04-27
Fixed issue related to allowing of insecure traffic

arzzon:
**Description**:  Fixed issue of CIS allowing insecure traffic for all the routes when one route is having insecureEdgeTerminationPolicy policy as allow.

**Changes Proposed in PR**: Skip adding forwarding policy rule to HTTP virtual server if the route doesn't allow insecure traffic.

## General Checklist
- [ ] Smoke testing completed

### DIFF
--- a/pkg/controller/nativeResourceWorker_test.go
+++ b/pkg/controller/nativeResourceWorker_test.go
@@ -246,6 +246,93 @@ var _ = Describe("Routes", func() {
 			Expect(len(rsCfg.Monitors)).To(BeEquivalentTo(2))
 
 		})
+		It("Checks whether Forwarding policy is added correctly", func() {
+			routeGroup := "default"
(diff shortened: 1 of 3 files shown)


pkg/controller/nativeResourceWorker_test.go
```diff
@@ -160,6 +160,93 @@ var _ = Describe("Routes", func() {
 			mockCtlr.removeDeletedRouteFromHostPathMap()
 			Expect(len(mockCtlr.processedHostPath.processedHostPathMap)).To(BeEquivalentTo(0))
 		})
+		It("Checks whether Forwarding policy is added correctly", func() {
+			routeGroup := "rg1"
```
vklohiya, line 164: routeGroup should be same as route's namespace.

PR #2392, opened 2022-05-04
Editing error log for local ConfigMap

Vidhi-Pat:
**Description**:  Editing error log for local ConfigMap

**Changes Proposed in PR**: Editing error log for local ConfigMap

### DIFF
--- a/pkg/controller/nativeResourceWorker.go
+++ b/pkg/controller/nativeResourceWorker.go
@@ -678,7 +678,7 @@ func (ctlr *Controller) processConfigMap(cm *v1.ConfigMap, isDelete bool) (error
 	} else if len(es.ExtendedRouteGroupConfigs) > 0 {
 		ergc := es.ExtendedRouteGroupConfigs[0]
 		if ergc.Namespace != cm.Namespace {
-			return fmt.Errorf("Invalid Extended Route Spec Block in configmap: %v/%v", cm.Namespace, cm.Name), true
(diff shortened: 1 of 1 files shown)


pkg/controller/nativeResourceWorker.go
```diff
@@ -678,7 +678,7 @@ func (ctlr *Controller) processConfigMap(cm *v1.ConfigMap, isDelete bool) (error
 	} else if len(es.ExtendedRouteGroupConfigs) > 0 {
 		ergc := es.ExtendedRouteGroupConfigs[0]
```
sravyap135, line 679: Need to log if there is more than one entry in local ConfigMap as well.

pkg/controller/nativeResourceWorker.go
```diff
@@ -678,7 +678,7 @@ func (ctlr *Controller) processConfigMap(cm *v1.ConfigMap, isDelete bool) (error
...
 		if ergc.Namespace != cm.Namespace {
-			return fmt.Errorf("Invalid Extended Route Spec Block in configmap: %v/%v", cm.Namespace, cm.Name), true
+			return fmt.Errorf("Invalid Extended Route Spec Block in configmap: Multiple namespaces found in %v/%v", cm.Namespace, cm.Name), true
```
sravyap135, line 681: Vidhi, can we log to mention the correct namespace at index 0 - as users are free to use multiple namespaces

sravyap135 (approved):

PR #2411, opened 2022-05-16
Performance improvement while processing endpoint events

sravyap135:
Signed-off-by: Sravya Pondugula <sravyap135@gmail.com>

**Description**: Performance improvement while processing endpoint events

**Changes Proposed in PR**:
- maintain service to resource name map for VS, TS, IngressLink, ServiceTypeLB resources
- Handles create, update and delete events. When a CRD pool backend service is updated, we initiate a delete + create event. 

- [x] Smoke testing completed


vklohiya (approved):
LGTM

PR #2414, opened 2022-05-17
exclude core service pods from NPL processing

lavanya-f5:
**Description**:  exclude core service pods from NPL processing
## General Checklist
- [x] Smoke testing completed


sravyap135 (approved):
LGTM

PR #2423, opened 2022-05-26
fix for service with named ports

### LINKED ISSUE #2294: Specifying service port by name instead of port number causing pool with 0 members
#### Setup Details
 CIS Version : 2.7.1
 BIGIP Version: Big IP 1.,1
 AS3 Version: 3.25      
 Agent Mode: AS3       
 Orchestration: K8S     
 Orchestration Version:       
 Pool Mode: Nodeport      
 Additional Setup details: Rancher being used.

#### Description  
Submitting on behalf of a customer. I think this is known by PM but I couldn't find an existing github issue, so I'd like to be able to track any plans to support, or official response from PM if support is not planned.

It appears that named ports within a service are not supported, and actual port numbers are required.

I'm no expert but as you probably already know, it looks like there's a way to [discover the port number](https://kubernetes.io/docs/concepts/services-networking/service/#dns) via DNS. 

#### Steps To Reproduce
If I have a service to with named ports (eg, targetPort: insecure-port and targetPort: secure-port), then CIS does not work as expected.
If I change the service to use targetPort: 80 and targetPort: 443 instead of the named aliases, then CIS works correctly.

Please let us know if there's any plan to support, and thanks!

Vidhi-Pat:
**Description**:   Fix for service with named ports and NAT ports with service TypeLB and NodePort

**Changes Proposed in PR**: Fix for service with named ports and NAT ports with service TypeLB and NodePort

**Fixes**: resolves #2294

## General Checklist
- [x] Smoke testing completed

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -24,6 +24,7 @@ Bug Fixes
 * :issues:`2323` Fixed file and examples links in ingresslink document
 * :issues:`2151` Fix for adding unique pool members only to AS3 declaration with AS3 configmap
 * :issues:`2326` Support for Configmap resource with NodePortLocal mode
+* :issues:`2294` Fix for service supporting with named ports
 
 2.8.1
 -------------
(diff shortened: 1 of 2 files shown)


vklohiya (approved):
LGTM

PR #2424, opened 2022-05-30
Extended support for snat in policy for VS and TS

arzzon:
**Description**:  Extended support for snat in policy for VS and TS

**Changes Proposed in PR**: Extended support for snat in policy for VS and TS


vklohiya (approved):
LGTM

PR #2436, opened 2022-06-06
Fix for map update when route is deleted

vklohiya:

**Description**: Fix for map update when route is deleted

## General Checklist

- [x] Smoke testing completed


subbuv26 (approved):
LGTM

PR #2456, opened 2022-06-14
Enqueue update event for resources during node update

sravyap135:
**Description**:  Enqueue update event for resources during node update

**Changes Proposed in PR**:
- Previously, create event was getting enqueued, and since the resource is already processed, the resource wasn't getting processed. Fixed to include an update event while enqueuing


nandakishorepeddi (approved):
LGTM

PR #2467, opened 2022-06-27
Removing the additional call for flags usage

vklohiya:

**Description**: Removing the additional call for flags usage

## General Checklist
- [ ] Smoke testing completed


subbuv26 (approved):
LGTM

PR #2500, opened 2022-07-15
Data sent by teems is stopped in case of network failure

vklohiya:

**Description**:  Data sent by teems is stopped along with retries in case of network failure 

## General Checklist
- [x] Smoke testing completed


lavanya-f5 (approved):
LGTM

PR #2510, opened 2022-07-25
Alias Port support in monitor

### LINKED ISSUE #2361: [CRD] Allow monitoring of an alias port in VirtualServer and TransportServer
#### Title
Allow monitoring of an alias port in VirtualServer and TransportServer

#### Description

As an app owner, I want to be able to monitor an alias port that is different from the port that is used to serve traffic. 
I am using calico BGP (clusterIP mode) integration so will LB directly to pods. 

#### Actual Problem
Today I cannot specify an aliasPort (a.k.a. targetPort) in the CRDs that makes the monitor target a different port to determine the health of a pool member. 

#### Solution Proposed

In the TransportServer CRD and VirtualServer CRD monitor block, add a parameter like targetPort to define the monitored port. 

- In BIG-IP TMUI, this translates to "Alias Service Port" on BIG-IP pool. 
- In AS3, this translates to "targetPort" in the Monitor class. 

#### Alternatives
the only alternative is to use AS3 configmaps, but today we cannot use both configmaps and CRDs so will be high impact.

lavanya-f5:
**Description**:  Alias Port support in monitor
**Changes Proposed in PR**:

**Fixes**: resolves #2361

## General Checklist
- [x] Smoke testing completed

## CRD Checklist

### DIFF
--- a/config/apis/cis/v1/types.go
+++ b/config/apis/cis/v1/types.go
@@ -73,11 +73,12 @@ type Pool struct {
 
 // Monitor defines a monitor object in BIG-IP.
 type Monitor struct {
-	Type     string `json:"type"`
-	Send     string `json:"send"`
-	Recv     string `json:"recv"`
-	Interval int    `json:"interval"`
-	Timeout  int    `json:"timeout"`
+	Type       string `json:"type"`
+	Send       string `json:"send"`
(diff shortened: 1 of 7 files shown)


nandakishorepeddi (approved):
LGTM

PR #2511, opened 2022-07-25
Added support for netmask for Virtual server

### LINKED ISSUE #1850: Add support for configuring `mask` for a Virtual Server
#### Title
Add support for configuring `mask` for a Virtual Server

#### Description
Currently, the controller does not support having a mask or a IP in CIDR notation for a virtual server.  We have a use case where we have HA F5 devices and want a virtual server to have a destination of something like `xx.xx.xx.xx/31`.

#### Actual Problem
I created an ingress that has the following annotation:
```
virtual-server.f5.com/ip: 10.130.11.8/31
```
but after applying it to the cluster, the controller throws the following:
```
2021/05/27 19:29:25 [WARNING] Error when setting Ingress status IP for virtual server ingress_10-130-11-8-31_443: Ingress.extensions "f5-ingress" is invalid: status.loadBalancer.ingress[0].ip: Invalid value: "10.130.11.8/31": must be a valid IP address
```
It seems like CIDR notation for the `virtual-server.f5.com/ip` annotation is not supported

#### Solution Proposed
Looking at the actual configuration on the F5, it seems like the configuration of something with a destination of `x.x.x.x/31` on the frontend has the following fields:
```
destination **x.x.x.x:https**
mask 255.255.255.254
```
Referring to the [BIGIP documentation](https://clouddocs.f5.com/api/icontrol-rest/APIRef_tm_ltm_virtual.html), `mask` is indeed the field that should be configured.

Therefore, I am proposing either
1) CIDR notation support is added to the `virtual-server.f5.com/ip` annotation OR
2) Add a new annotation (e.g. `virtual-server.f5.com/mask`) that would be used to configure the mask directly.

#### Alternatives
N/A

#### Additional context
**Current versions (in case it matters):**
Kubernetes: 1.18.12
`k8s-bigip-ctlr`: 1.8.1
BIGIP F5 device: 14.1.4 Build 0.0.11 Final

I know that we're quite behind on the `k8s-bigip-ctlr` version, but looking through the code, I don't think the newer version supports having a mask either (please correct me if I'm wrong!).

arzzon:
**Description**:  Added support to configure netmask for Virtual Server for Ingress

**Changes Proposed in PR**: CIDR notation is allowed for virtual-server.f5.com/ip in the annotation of ingress resources.

**Fixes**: resolves #1850

## General Checklist
- [x] Smoke testing completed

## CRD Checklist

### DIFF
--- a/cmd/k8s-bigip-ctlr/main.go
+++ b/cmd/k8s-bigip-ctlr/main.go
@@ -1185,6 +1185,7 @@ func getAppManagerParams() appmanager.Params {
 		ProcessAgentLabels:     getProcessAgentLabelFunc(),
 		DefaultRouteDomain:     *defaultRouteDomain,
 		PoolMemberType:         *poolMemberType,
+		Agent:                  *agent,
 	}
 }
 
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -9,7 +9,7 @@ Added Functionality
 * Added Support for AS3 3.38.0
 * CRD:
         * AllowSourceRange support for VirtualServer CRs and Policy CRs. See `Examples <https://github.com/F5Networks/k8s-bigip-ctlr/tree/master/docs/config_examples/customResource/>`_
-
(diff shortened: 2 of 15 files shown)


vklohiya (approved):
LGTM

vklohiya (approved):
LGTM

PR #2532, opened 2022-08-08
Update poolmembers in nodeport mode only when there's node update

lavanya-f5:
**Description**:  Update poolmembers in nodeport mode only when there's node update

- [x] Smoke testing completed


vklohiya (approved):
LGTM

subbuv26 (approved):
LGTM

PR #2539, opened 2022-08-11
Exclude prometheus services from core services

### LINKED ISSUE #2325: Supporting Prometheus service in CRDs (similar to 2280)
Similar to the Bug 2280, Prometheus services cannot be published through CRDs. They are part of the core services and CIS (CRD mode) doesn't discover any services. 
Please remove Prometheus-k8s as part of the core services as done with Grafana (see the code below). With CIS Ingress services it works fine.

image

lavanya-f5:
**Description**:  Exclude prometheus services from core services
**Fixes**: resolves #2325 

## General Checklist


vklohiya (approved):
LGTM

PR #2544, opened 2022-08-16
Remove invalid config from retry tenant map

lavanya-f5:
**Description**:  Currently, when AS3 returns 422 we keep retrying every 30s for failed tenants. So even if config is updated to correct reference, currently CIS not processing the new update.

**Changes Proposed in PR**:
   422 Unprocessable Entity response status code indicates invalid config.Client should not repeat this request without modification.So remove the tenant config from retry map.

## General Checklist
- [x] Smoke testing completed

### DIFF
--- a/pkg/controller/backend.go
+++ b/pkg/controller/backend.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"encoding/json"
 	"fmt"
+	"net/http"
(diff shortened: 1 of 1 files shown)


nandakishorepeddi (approved):
LGTM

PR #2563, opened 2022-08-30
Enable share nodes with multi-partition support only

vklohiya:

**Description**: Enable share nodes with multi-partition support only

## General Checklist
- [x] Smoke testing completed


charanm08 (commented):
LGTM

charanm08 (approved):

PR #2608, opened 2022-10-12
Support for grouping vs by hg across namespaces

### LINKED ISSUE #2469: multiple VS with same IP address in different namespaces not working
#### Setup Details
 CIS Version : 2.8.1      
 BIGIP Version: Big IP 16.1.3      
 AS3 Version: 3.25
 Agent Mode: AS3
 Orchestration: K8S

#### Description  
Multiple VS's that share same IP address using hostGroup parameter will not work if they are in different namespaces.

#### Steps To Reproduce
1) create 2x namespaces
2) create a VS crd in namespace 1, with IP address and hostGroup parameter.
3) create another VS crd in namespace 2, with matching IP address, virtualServerName, and hostGroup parameter.

#### Expected Result 
A single VS is created on BIG-IP, where the HTTP policy contains rules for hostnames of both VS crd's.

#### Actual Result  
The first VS is "overwritten" by the second VS, so config of first VS crd is lost from BIG-IP.

#### Observations
Customer hit this issue, and I was able to replicate. It appears that our hostGroup parameter only works if the 2x VS crd's are in the same namespace.

Can we support a single Virtual IP address across multiple namespaces? 

Customer has multiple applications and would like to use a single IP address, but their apps will be in different namespaces. The alternative would be to manage multiple IP's which would be cumbersome for them, or to use IPAM and then DNS automation, which is more work for them.

lavanya-f5:
**Description**:  Support for grouping vs by hg across namespaces
**Fixes**: resolves #2469 
## General Checklist
- [x] Smoke testing completed

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -11,7 +11,7 @@ Added Functionality
     * Support for health monitors using route annotations See `Examples <https://github.com/F5Networks/k8s-bigip-ctlr/tree/master/docs/config_examples/next-gen-routes/routes>`_
     * CIS configures GTM configuration in default partition
     * Pool reselect support for VS and TS
-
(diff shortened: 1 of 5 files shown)


nandakishorepeddi (approved):
LGTM

vklohiya (approved):
LGTM

PR #2616, opened 2022-10-18
Support to provide the same VIP for TS and VS CRs using hostGroup

### LINKED ISSUE #2581: IPAM to provide the same IP for different TS
#### Title
IPAM to provide the same IP for different TS or VS

#### Actual Problem
There are many applications that are using multiple port (like 80,443). To publish them on BIGIP we would need 2 TS or VS with IPAM. When using either TS or VS with IPAM we cannot publish them under the same IP address. IPAM will consider 2 different TS/VS resource and provide 2 different IPs. These makes it impossible to publish the application.

#### Solution Proposed
There has to be a label that would force the indicate to IPAM that it should provide the same IP

nandakishorepeddi:
**Description**:  Support to provide the same VIP for TS and VS CRs using hostGroup

**Changes Proposed in PR**:

**Fixes**: resolves #2581 

## General Checklist
- [x] Smoke testing completed

## CRD Checklist


vklohiya (approved):
LGTM

lavanya-f5 (approved):
LGTM

PR #2641, opened 2022-11-15
Add support for recv property in https monitor

vklohiya:

**Description**:  Add support for recv property in https monitor

## General Checklist
- [x] Smoke testing completed


charanm08 (approved):
LGTM

PR #2710, opened 2023-01-11
Validate insecure Virtual Server CR

### LINKED ISSUE #2686: CIS is not populating DNS pool with HTTP virtual servers when httpTraffic: allow exists but TLSProfile does not exist
#### Setup Details
 **CIS Version : 2.8.0, 2.8.1, 2.9.0, 2.9.1, 2.10.0, 2.10.1, 2.11.0** (CIS 2.7.1 works fine)
 Build: f5networks/k8s-bigip-ctlr     
 BIGIP Version: Big IP 16.1.3.2     
 AS3 Version: 3.25      
 Agent Mode: AS3       
 Orchestration: K8S        
 Pool Mode: Cluster      

#### Description  

If a HTTP VS exists with ```httpTraffic: allow``` but does not specify a ```TLSProfileName```, the HTTP VS is created on BIG-IP but not added to any GTM pool.

This was working fine in CIS v2.7.1. Customer and me have tested all versions 2.8.0 through 2.11.0 and this problem exists. I believe the fix for #2212 introduced this behavior in v 2.8.0.

```httpTraffic``` attribute should probably not be used on a insecure VS that does not have any TLSProfile referenced. However, this attribute is allowed which caused this problem.

My suggestion: If the ```httpTraffic``` attribute value is ```allow```, but there is no ```tlsProfileName``` attribute, then we could either:
a)	Throw an error to force the user to remove this line, or
b)	Allow for this condition and still create a GTM pool member.

#### Steps To Reproduce
1) Deploy CIS version 2.8.0 or higher with arguments for F5 GTM server
2) Deploy insecure VS and EDNS resources where the VS has httpTraffic: allow (examples below)
3) With CIS version 2.7.1 the LTM VS listening on tcp/80 is added to GTM pool. However in versions 2.8.0 thru latest (2.11.0), the GTM pool is empty.

````yaml
apiVersion: "cis.f5.com/v1"
kind: VirtualServer
metadata:
  name: hello-world-virtual-server
  namespace: nginx-ingress
  labels:
    f5cr: "true"
spec:
  host: demo.my-f5.com
  tlsProfileName: hello-world-tls  # --> This will attach hello-world-tls TLSProfile
  virtualServerAddress: "20.65.20.104"
  virtualServerName: "hello-world-vs"
  **httpTraffic: allow**    # this line is what causes the problem with populating GTM pool
  pools:
  - path: /
    service: nginx-ingress
    servicePort: 80
    monitor:
      type: http
      interval: 10
      timeout: 31
      send: "/"
      recv: ""
````

````yaml
apiVersion: cis.f5.com/v1
kind: TLSProfile
metadata:
  name: hello-world-tls
  namespace: nginx-ingress
  labels:
    f5cr: "true"
spec:
  tls:
    termination: edge
    clientSSL: /Common/clientssl
    reference: bigip             
  hosts:
    - demo.my-f5.com
````

#### Expected Result 

VirtualServer is a member of GTM pool

#### Actual Result  

GTM pool is empty.

#### Workaround
remove ```httpTraffic: allow``` from any VS that is not secured by TLS and requires ExternalDNS.

arzzon:
**Description**:  Validate insecure Virtual Server CR

**Changes Proposed in PR**: Added check to ensure HTTPTraffic is not set for insecure VS

**Fixes**: resolves #2686

## General Checklist
- [x] Smoke testing completed

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -7,6 +7,7 @@ Next Release
 Added Functionality
 ```````````````````
 * `Issue 2682 <https://github.com/F5Networks/k8s-bigip-ctlr/issues/2682>`_: Support to Enable "HTTP MRF Router" on VirtualServer CRD required for HTTP2 Full Proxy feature
+* `Issue 2686 <https://github.com/F5Networks/k8s-bigip-ctlr/issues/2686>`_: Validate insecure Virtual Server CR
 
 Bug Fixes
 `````````
--- a/pkg/controller/validate.go
+++ b/pkg/controller/validate.go
@@ -42,6 +42,12 @@ func (ctlr *Controller) checkValidVirtualServer(
 		log.Infof("VirtualServer %s is invalid", vsName)
 		return false
 	}
+	// Check if HTTPTraffic is set for insecure VS
+	if vsResource.Spec.TLSProfileName == "" && vsResource.Spec.HTTPTraffic != "" {
+		log.Errorf("HTTPTraffic not allowed to be set for insecure VirtualServer: %v", vsName)
+		return false
+	}
+
 	bindAddr := vsResource.Spec.VirtualServerAddress
 	if ctlr.ipamCli == nil {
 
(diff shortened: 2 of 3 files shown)


lavanya-f5 (approved):
LGTM

PR #2727, opened 2023-01-17
multiple vip support for vs

### LINKED ISSUE #2666: Feature Request for multiple Virtual Address's in a single CRD
#### Title
Looking to have a single CRD definition for multiple virtual addresses, since cloud deployments could need multiple private addresses this would solve the need for two virtualservers deployed.

#### Description
I assumed reading the documentation that service_address would solve this, but looking into the schema this appears to just be an additional setting to the single virtual address. If this is actually possible, I have not been able to find an example on any F5 property. 

#### Actual Problem
Maintaining two configuration files for one object is not a great way

#### Solution Proposed
AS3 accepts this as a string of addresses, that in turn become multiple virtual servers. CRD should be able to insert the same when crafting the as3 declaration

Something like:

```
apiVersion: cis.f5.com/v1
kind: VirtualServer
metadata:
  name: bigip-virtualserver-nginx
  namespace: nginx-ingress
  labels:
    f5cr: "true"
spec:
  virtualServerName: "ingresslink-calalang-net"
  serviceAddress: [{10.0.2.50, 10.0.2.51}]
  virtualServerHTTPSPort: 443
  tlsProfileName: bigip-calalang-net-tls
  httpTraffic: redirect
  pools:
    - monitor:
        interval: 20
        recv: ""
        send: "GET /\r\n"
        timeout: 10
        type: http
      path: /
      service: nginx-ingress
      servicePort: 80
```

#### Alternatives
Creating multiple CRD resources

#### Additional context
I thought maybe this class in CRD could do what I'm looking because the docs are worded that way. However, it only appears to just be an additional feature for the single address

https://clouddocs.f5.com/containers/latest/userguide/crd/virtualserver.html?highlight=serviceaddress

image

lavanya-f5:
**Description**:   multiple vip support for vs CRD

**Fixes**: resolves #2666

## General Checklist
- [x] Smoke testing completed

## CRD Checklist


nandakishorepeddi (approved):
LGTM

vklohiya (approved):
LGTM

PR #2758, opened 2023-02-07
fix ltm policy for default port

lavanya-f5:
**Description**:  With default http or https port i.e 80 or 443, the host header could be sent as just the host or host:port from client. Fixed ltm policy created by cis to support this.

## General Checklist
- [x] Smoke testing completed


nandakishorepeddi (approved):
LGTM

vklohiya (approved):
LGTM

PR #2761, opened 2023-02-08
fix named target port nodeport

lavanya-f5:
**Description**:  fix named target port with nodeport

- [x] Smoke testing completed


vklohiya (approved):
LGTM

PR #2802, opened 2023-03-09
Fix for 422 error when default partition deleted

vklohiya:
**Description**:  Fix for 422 error when default partition deleted

## General Checklist
- [ ] Smoke testing completed


arzzon (approved):
LGTM

PR #2815, opened 2023-03-16
fix host check legacy routes

lavanya-f5:
**Description**:  fix host check legacy routes

## General Checklist
- [x] Smoke testing completed


nandakishorepeddi (approved):
LGTM

PR #2855, opened 2023-04-20
Multi Cluster service support with improved node & service update processing

vklohiya:
**Description**:  Multi Cluster service support with improved nodes & service update processing 

**Changes Proposed in PR**: This PR addresses following:

1.  **The initial processing of CIS resources:** Earlier CIS used to process all the services at CIS start up and create a poolMember cache. With this changes now CIS does not cache any services at startup instead it directly process the routes/serviceTypeLB/VS/TS/IngressLink and then maintains the pool member cache only for services which are referenced in these resource. This will improve the CIS start up time as we won't process all the services in system and also improve memory used by CIS as less services will be in the cache.

2. **Processing of service updates** Earlier for service changes we use to have svcResourceCache which used to map service to a rsConfig, due to that even for a single service update we use to update all the pools for that particular rsConfig. With the new mappings clusterSvcMap we are able to identify the pools related to the service and update the pool members only for single pool. This also helps us to use a single function to use for all kind of resources(routes/serviceTypeLB/VS/TS/IngressLink) and types of member(NodePort/Cluster/NodePortLocal). 

3. **Node informers improvement:** Earlier we use to create the node informer for every namespace. Now we have centralised the node informer and only one informer is created for each cluster. This will improve memory used by CIS.

4. **Node Update Processing:** Earlier node updates were processed only for VS/TS/IL resources and we use to enqueue all these resources to resource queue one by one which resulted in longer resource processing as we were framing all the rsCfg  for these resources again instead of just updating the pool members. With the new change, if there is a node update we just update the pool members for all the rsCfg present in the CIS cache regardless of resource Type. This will result in fast processing of node update.

5. **VxLanMgr improvement:** Earlier we were creating a new vxLanMgr whenever a node update is received. With this PR we have centralised the VxLanMgr and only single vxlanMgr is created. This will improve memory used by CIS.

## General Checklist
- [x] Smoke testing completed

## CRD Checklist

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -144,6 +144,7 @@ func NewController(params Params) *Controller {
 		workqueue.DefaultControllerRateLimiter(), "nextgen-resource-controller")
 	ctlr.comInformers = make(map[string]*CommonInformer)
 	ctlr.multiClusterPoolInformers = make(map[string]map[string]*MultiClusterPoolInformer)
+	ctlr.multiClusterNodeInformers = make(map[string]*NodeInformer)
 	ctlr.nrInformers = make(map[string]*NRInformer)
 	ctlr.crInformers = make(map[string]*CRInformer)
 	ctlr.nsInformers = make(map[string]*NSInformer)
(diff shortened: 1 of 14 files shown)


lavanya-f5 (approved):
LGTM

charanm08 (approved):
LGTM

nandakishorepeddi (approved):
LGTM

PR #2874, opened 2023-05-04
Fix for handling fail-over scenarios

vklohiya:
**Description**:  Fix for handling fail-over scenarios

**Changes Proposed in PR**: With these changes we will we performing following actions at CIS startup so we delete any partition which is not monitored by CIS HA.

1. Get AS3 Declaration
2. Compare tenants got from BigIP AS3 declaration with current rsConfig tenants 
3. Any tenant which is not there in current rsConfig needs to be removed from BIGIP. 
4. Add an empty partition config in rsConfig to delete the tenant.

Note: We are using tenant label to identify that tenant is created by this CIS instance. Tenant label consist of BIGIP-IP and default partition to identify unique CIS instance.

## General Checklist
- [x] Smoke testing completed

### DIFF
--- a/pkg/controller/backend.go
+++ b/pkg/controller/backend.go
@@ -218,6 +218,27 @@ func (agent *Agent) PostConfig(rsConfig ResourceConfigRequest) {
 	}
 }
 
+// removeDeletedTenantsForBigIP will check the tenant exists on bigip or not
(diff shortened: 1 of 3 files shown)


charanm08 (approved):
LGTM

nandakishorepeddi (approved):
LGTM

arzzon (approved):
LGTM

PR #2878, opened 2023-05-08
Add HTTP Monitor Support for Transport Server CR

### LINKED ISSUE #2550: Ability to specify monitors for TransportServer
#### Description
Right now you can only specify some basic monitor settings on `TransportServer`. You should be able to configure things like HTTP checks or other custom monitor settings from the CRD.

#### Actual Problem
If you use layer 4 with a transportserver to send traffic to an ingress controller because you need to terminate TLS on your cluster, then you cannot setup health checks other than a basic "port open" check.

#### Solution Proposed
Ability to configure more details of the monitor like the parent being an HTTP check instead of a basic port open check.

#### Context
Ingress controllers like Istio allow you to do a HTTP check check on an endpoint to see if the pod is shutting down and should not accept new traffic for a graceful upgrade. Without F5 checking this endpoint, it won't know to stop sending traffic until the whole pod is killed.

nandakishorepeddi:
**Description**:  Add HTTP Monitor Support for Transport Server CR

**Changes Proposed in PR**:
Add HTTP Monitor Support for Transport Server CR

**Fixes**: resolves #2550

## General Checklist
- [ ] Smoke testing completed

## CRD Checklist

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -25,13 +25,16 @@ Added Functionality
         * Add Support of ServerSide HTTP2 Profile in Policy CR and VS CR
         * Support setting http mrf router option from policy CR(applicable only to VS cr)
         * Support for setting http analytics profile from policy CR
+        * Support HTTP Monitor for Transport Server CR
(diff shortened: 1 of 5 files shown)


charanm08 (approved):
LGTM

PR #2933, opened 2023-06-15
fix nodeport cccl not posting config

lavanya-f5:
**Description**:  fix nodeport cccl not posting config


nandakishorepeddi (approved):
LGTM

PR #2944, opened 2023-06-22
Handle scenario where HA cluster used as external cluster

arzzon:
**Description**:  Handle scenario where HA cluster used as external cluster

**Changes Proposed in PR**: Handle scenario where HA cluster used as external cluster

## General Checklist
- [x] Smoke testing completed


nandakishorepeddi (approved):
LGTM

PR #2991, opened 2023-07-19
fix static routes creation on cis restart with mc

lavanya-f5:
**Description**:  fix static routes creation on cis restart with mc

## General Checklist
- [x] Smoke testing completed


nandakishorepeddi (approved):
LGTM

PR #2992, opened 2023-07-20
support reading CNI data from orchestration-cni arg

lavanya-f5:
**Description**:  support reading CNI data from orchestration-cni arg


nandakishorepeddi (approved):
LGTM

PR #2995, opened 2023-07-25
Fix handling service creation after VS/TS CR creation

arzzon:
**Description**:  Fix handling service creation after VS/TS CR creation

**Changes Proposed in PR**: Reprocess VS/TS resources associated with the service when is created the service is created after the creation of the VS/TS.

## General Checklist
- [x] Smoke testing completed


vklohiya (approved):
LGTM

PR #3003, opened 2023-08-01
Rename multiCluster parameters and update documentation

### LINKED ISSUE #2990: multicluster preview: naming suggestions
#### Title

multicluster preview: naming suggestions

#### Description

- Never use the word cluster standalone because it can refer to:
  * CIS cluster
  * BIG-IP cluster
  * OpenShift cluster

and we should name as above to avoid ambiguity

- If https://github.com/F5Networks/k8s-bigip-ctlr/issues/2989 is not going to be implemented in the short term mode values should be renamed:

    active: really means active-active
    standby: really means active-standby
    
and should just be named active-active or active-standby to reflects better the actual behavior

- I find confusing the the use of primaryCluster and secondaryCluster  as per above, they they are really referring to CIS. I find that primary and secondary CIS should be an attribute of the cluster indicating whether the cluster is running CIS and in which mode.

- The global ConfigMap should be the same regardless of where it is appplied and at present localClusterRatio makes it not possible

- I would propose to replace the `--cis-type` parameter with a `--cluster-name` parameter. CIS would just the figure out if it is HA role according to the ConfigMap.

- Overall the ConfigMap could look as follows:

```
extendedSpec: |
  HighAvailabilityCIS:
    primaryEndPoint: http://10.145.72.114:8001
    probeInterval: 30
    retryInterval: 3
  multiClusterConfigs:
  - clusterName: cluster1
    secret: default/kubeconfig1
    default-mode: active
    primaryCIS: true
  - clusterName: cluster2
    secret: default/kubeconfig2
    default-mode: standby
    secondaryCIS: true
  - clusterName: cluster3
    secret: default/kubeconfig3
    default-mode: active
  - clusterName: cluster4
    default-mode: standby
    secret: default/kubeconfig4
```
Please note the next field renaming following the previous comments:

- `highAvailabilityClusterConfigs` has been renamed to `HighAvailabilityCIS` to indicate explicitly we are referring to CIS HA only
- `primaryClusterEndPoint` has been renamed to `primaryEndPoint`

arzzon:
**Description**:  Rename multiCluster parameters and update documentation

**Changes Proposed in PR**:  Following changes are done in this PR:
1. HA mode names and documentation have been improved
2. HA config and endpoint parameters have been renamed
3. MultiCluster documentation has been improved

**Fixes**: resolves #2990

## General Checklist
- [x] Smoke testing completed

## CRD Checklist

### DIFF
--- a/docs/config_examples/multicluster/README.md
+++ b/docs/config_examples/multicluster/README.md
@@ -1,4 +1,4 @@
-# OpenShift Multi-Cluster Preview
+# OpenShift/Kubernetes Multi-Cluster Preview
 
 This page documents a new feature in CIS for Multi-Cluster. This is a preview release which supports limited features and is not recommended to use in production environments. To provide feedback on Container Ingress Services or this documentation, please file a [GitHub Issue](https://github.com/F5Networks/k8s-bigip-ctlr/issues)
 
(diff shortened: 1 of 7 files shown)


vklohiya (approved):
LGTM

PR #3006, opened 2023-08-04
dont push cccl config when cis is in secondary mode

lavanya-f5:
**Description**:  dont push cccl config when cis is in secondary mode 

## General Checklist
- [x] Smoke testing completed


vklohiya (approved):
LGTM

PR #3009, opened 2023-08-08
Renaming the cis-type to multi-cluster-mode

vklohiya:
**Description**:  Fix for node removal when app is scaled down to zero in non multi cluster mode

## General Checklist
- [ ] Smoke testing completed


vklohiya:
Closing the PR as we will create separate Pr for each commit.

PR #3010, opened 2023-08-09
Fix for node removal when app is scaled down to zero in non multi cluster mode

vklohiya:
**Description**:  Fix for node removal when app is scaled down to zero in non multi cluster mode

## General Checklist
- [x] Smoke testing completed


nandakishorepeddi (approved):
LGTM

PR #3044, opened 2023-08-24
Fix for AB deployment while using insecure termination with routes

vklohiya:
**Description**:  Fix for AB deployment while using insecure termination with routes

## General Checklist
- [x] Smoke testing completed


nandakishorepeddi (approved):
LGTM

PR #3045, opened 2023-08-24
Fix for AB deployment while using insecure termination with VS CR

vklohiya:
**Description**:  Fix for AB deployment while using insecure termination with VS CR

## General Checklist
- [x] Smoke testing completed


arzzon (approved):
LGTM

PR #3046, opened 2023-08-24
Improvements in pod liveness probe based health monitor for NextGenRoutes

### LINKED ISSUE #3041: livenessprobes monitors create incorrect probes
#### Setup Details
 CIS Version : 2.X      
 Build: quay.io/f5networks/k8s-bigip-ctlr-devel:c20b620d3e972dd928f215552671bcb97b0b8eea       

#### Description  

At present the following livenessProbe:

```
        livenessProbe:
          httpGet:
            path: /
            port: 3000
          initialDelaySeconds: 5
          periodSeconds: 4
          timeoutSeconds: 1
          successThreshold: 1
          failureThreshold: 3  
```

CIS configures the following:

```
ltm monitor http Shared/guestbook_app1_3000_eng_caas_guestbook_app1_monitor {
    adaptive disabled
    adaptive-limit 1000
    adaptive-sampling-timespan 180
    defaults-from /Common/http
    destination *:*
    interval 1
    ip-dscp 0
    partition OpenShift-MultiCluster
    recv none
    recv-disable none
    send "GET /\r\n"
    time-until-up 0
    timeout 2
}
```

This means that we send a non HTTP compliant request, technically it is an HTTP 0.9 request. This actually triggers the following response from the pool member:

```
HTTP/1.1 400 Bad Request
Content-Type: text/plain; charset=utf-8
Connection: close
```

Additionally, please note that we setup an interval of 1 and a timeout of 2 but the customer defined the following periodSeconds: 4. 

I also think that we should translate kubernetes initialDelaySeconds: 5 into BIG-IP's "time until up" parameter.

arzzon:
**Description**:  Improvements in pod liveliness probe based health monitor for NextGenRoutes

**Changes Proposed in PR**: 
 Improvements in pod liveness probe based health monitor for NextGenRoutes:
             * Use request conforming to HTTP/1.0 protocol in send string
            * Update recv string to accept response code between 200 to 399
            * Set timeUntilUp using initialDelaySeconds from pod liveliness probe
            * Use path field from pod livenessprobe in send string of health monitor
            * Add autoMonitor and autoMonitorTimeout option in baseRouteSpec

**Fixes**: resolves #3041 

## General Checklist
- [x] Smoke testing completed

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -8,6 +8,8 @@ Added Functionality
 ```````````````````
 **What's new:**
     * Multi Cluster
+    * Next Generation Routes
+        * Improvements in pod liveness probe based health monitor and autoMonitor support for NextGenRoutes. See `Example <https://github.com/F5Networks/k8s-bigip-ctlr/tree/master/docs/config_examples/next-gen-routes/configmap>`_
     * CRD
     * Support for AS3 GTM agent with separate GTM server
 Bug Fixes
--- a/docs/config_examples/next-gen-routes/README.md
+++ b/docs/config_examples/next-gen-routes/README.md
@@ -94,6 +94,8 @@ If it's specified in both the places then allow source range in policy CR has mo
 ### Support for Health Monitors from pod liveness probe
 CIS uses the liveness probe of the pods to form the health monitors, whenever health annotations not provided in the route annotations. 
 
+This behaviour can be changed by setting autoMonitor in baseRouteSpec of the extended configmap.
+
 ## Migration Guide
 Follow this for easy migration [Migration Guide](https://github.com/F5Networks/k8s-bigip-ctlr/blob/master/docs/config_examples/next-gen-routes/migration-guide.md)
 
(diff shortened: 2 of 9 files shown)


nandakishorepeddi (approved):
LGTM

lavanya-f5 (approved):
LGTM

vklohiya (approved):
LGTM

alonsocamaro:
We would like that you reconsider:

1. Customer remarked that liveness probes are meant to restart the containers, not to check if the containers are ready to accept traffic. 

Usually the probes used in both readiness and liveness  probes are the same but it might not be always the case. Hence:

- we ask to replace the liveness probes by readiness probes (and replace the keywords in the configs)

2. We previously understood that there RFE for liveness probes came from a different customer and hence it required to have it by default. Given that it is not the case we strongly encourage you to:

- make none the default value because readiness/liveness probes are a) a change in behaviour from all other previous CIS implementations and b) it might cause high load in customer systems. This customer, Kostas and me also think this is the most sensible approach. It is better to amend this now than further later on.

#### Solution Proposed

- change the behaviour to none ASAP
- replace liveness-probes to readiness-probes later on

#### Additional information

Note that Kubernete's NodePort also makes use of Readyness to send traffic to a POD or not:

https://kubernetes.io/docs/concepts/services-networking/service/

For a node port Service, Kubernetes additionally allocates a port (TCP, UDP or SCTP to match the protocol of the Service). Every node in the cluster configures itself to listen on that assigned port and to forward traffic to one of the ready endpoints associated with that Service. You'll be able to contact the type: NodePort Service, from outside the cluster, by connecting to any node using the appropriate protocol (for example: TCP), and the appropriate port (as assigned to that Service).

I find that it should be fixed before customers start using the keyword "liveness-probe" in their manifests

PR #3049, opened 2023-08-25
Fix for pod caching in nodeportlocal mode

vklohiya:
**Description**:  CIS is not caching the pods in nlpstore, Hence no pool members are added in nodeportlocal mode.

## General Checklist

- [x] Smoke testing completed


charanm08 (approved):
LGTM

PR #3077, opened 2023-09-06
support pool settings in policy cr

### LINKED ISSUE #3057: Enhance availability on scale-down and rollout updates of deployments
#### Title
Enhance availability on scale-down and rollout updates of deployments

#### Description
When there is a scale down event or when there is a deployment update, existing POD members disappear. The BIG-IP and the Kubernetes status is never going to be in lock-step so it is expected that some requests will always be sent to disappearing Pods.

Running an ab -n 20000 -c 20 on an URL with 30 PODs behind, it can be seen ~2% of failed requests 

When using an irule to force a LB::reselect when LB_FAILED is triggered (this can be done only once from an irule) this is reduced to 0.5%. Being able to increment the retries further would reduce this to 0%.

#### Actual Problem

There are failures when sending requests to disappearing Pods when doing regular operations in a Kubernetes cluster (ie: scale-in, app re-deployment)

#### Solution Proposed

- support in the Policy CR the pool's reselect retries attribute.

- Please also include the related pool settings:
  - "Action On Service Down".
  - "Slow Ramp Time"

Although pool member reselection, action on service down and slow ramp time are per pool in BIG-IP, they are general enough to used in a per policy basis and avoid the use of annotations.

#### Additional information

The KB below suggest to use N-1 but I find that way to large in a Kubernetes environment as it would create a syn flood storm. Implementing a dynamic value which can be also be hinted by the customer and that adapts to large and small pool seems challenging. Maybe a static value would be more predictable and ultimately preferable to start with.

See https://my.f5.com/manage/s/article/K42101213 and https://my.f5.com/manage/s/article/K10386

lavanya-f5:
**Description**: support pool reselect settings in policy cr

**Fixes**: resolves #3057 

## General Checklist
- [x] Smoke testing completed

## CRD Checklist

### DIFF
--- a/config/apis/cis/v1/types.go
+++ b/config/apis/cis/v1/types.go
@@ -311,14 +311,21 @@ type ExternalDNSList struct {
 }
 
 type PolicySpec struct {
-	L7Policies  L7PolicySpec  `json:"l7Policies,omitempty"`
-	L3Policies  L3PolicySpec  `json:"l3Policies,omitempty"`
-	LtmPolicies LtmIRulesSpec `json:"ltmPolicies,omitempty"`
-	IRules      LtmIRulesSpec `json:"iRules,omitempty"`
-	IRuleList   []string      `json:"iRuleList,omitempty"`
-	Profiles    ProfileSpec   `json:"profiles,omitempty"`
-	SNAT        string        `json:"snat,omitempty"`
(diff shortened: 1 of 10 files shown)


nandakishorepeddi (approved):
LGTM

PR #3099, opened 2023-10-04
Added support for readiness based auto monitoring

### LINKED ISSUE #3089: RFE: liveness probes feedback
#### Title
RFE: liveness probes feedback

#### Description

We would like that you reconsider:

1. Customer remarked that liveness probes are meant to restart the containers, not to check if the containers are ready to accept traffic. 

Usually the probes used in both readiness and liveness  probes are the same but it might not be always the case. Hence:

- we ask to replace the liveness probes by readiness probes (and replace the keywords in the configs)

2. We previously understood that there RFE for liveness probes came from a different customer and hence it required to have it by default. Given that it is not the case we strongly encourage you to:

- make none the default value because readiness/liveness probes are a) a change in behaviour from all other previous CIS implementations and b) it might cause high load in customer systems. This customer, Kostas and me also think this is the most sensible approach. It is better to amend this now than further later on.

#### Solution Proposed

- change the behaviour to none ASAP
- replace liveness-probes to readiness-probes later on

vklohiya:
**Description**:  Add support for readiness based auto monitoring

**Changes Proposed in PR**:  Moved from pod liveness probe based health monitor to readiness probe based health monitor for autoMonitor. Added support for TCP probe with readiness.

**Fixes**: resolves #3089 

## General Checklist
- [x] Smoke testing completed

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -9,16 +9,21 @@ Added Functionality
 **What's new:**
     * Multi Cluster
     * Next Generation Routes
-        * Improvements in pod liveness probe based health monitor and autoMonitor support for NextGenRoutes. See `Example <https://github.com/F5Networks/k8s-bigip-ctlr/tree/master/docs/config_examples/next-gen-routes/configmap>`_
+        * Moved from pod liveness probe based health monitor to readiness probe based health monitor for autoMonitor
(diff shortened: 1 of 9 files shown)


charanm08 (approved):
LGTM

arzzon (approved):
LGTM

PR #3111, opened 2023-10-13
Added support for pod-concurrent-connections

### LINKED ISSUE #3086: RFE: Implement annotation equivalent to haproxy.router.openshift.io/pod-concurrent-connections
#### Title

Implement annotation equivalent to haproxy.router.openshift.io/pod-concurrent-connections 

#### Description

Customer wants to have this functionality at Route level.

The functionality is implemented with the following AS3 PoolMember attribute: connectionLimit. That is, the annotation value will need to be applied to all pool members.

#### Actual Problem

This functionality avoids backends' overload and it has been widely used in BIG-IP.  Implementing it allows customers to switch from OpenShift's default HA-Proxy to BIG-IP.

#### Solution Proposed

Create an annotation at Route level. Ideally, this annotation should be available to equivalent resource types for functionality: VirtualServer and Ingress but also TransportServer.

#### Additional context

Implementing this at Route level means that if a service is reused between two Routes the connection limit value might be overridden. The customer is fine with this behavior. The customer prefers to have this at Route level instead of Service level. 

If two Routes with different connection limits share the same service, the customer can always create a separated Service to avoid the limits to be overridden.

vklohiya:
**Description**:  Added support for pod-concurrent-connections

**Changes Proposed in PR**:

**Fixes**: resolves #3086 

## General Checklist
- [x] Smoke testing completed

## CRD Checklist

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -10,6 +10,7 @@ Added Functionality
     * Multi Cluster
     * Next Generation Routes
         * Moved from pod liveness probe based health monitor to readiness probe based health monitor for autoMonitor
+        * Support for new route annotation **virtual-server.f5.com/pod-concurrent-connections**, See `Example <https://github.com/F5Networks/k8s-bigip-ctlr/tree/master/docs/config_examples/next-gen-routes/routes>`_
     * CRD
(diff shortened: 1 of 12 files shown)


arzzon (approved):
LGTM

PR #3178, opened 2023-12-04
Add MultiCluster support for IPAM controller by prepending custom keys

### LINKED ISSUE #3160: Infoblox IPAM assigns the same IP to each of the two different clusters.
#### Title
Infoblox IPAM assigns the same IP to each of the two different clusters.

#### Description
1. There are two different clusters.
2. Two different clusters each have F5 CIS installed and are linked to the same F5 BIG-IP Appliance LTM.
3. F5 CIS in two different clusters are connected to the same Infoblox IPAM and have the same Label and CIDR.
4. In two different clusters, set the LB Type as follows.
-- Service Name is the same. (ex: nginx-svc)
-- Infoblox IPAM Label is the same.

In the above situation, what we want is for two different clusters to be assigned different VIPs and set up LB with different virtual addresses on the F5 BIG-IP Appliance.

#### Actual Problem
If you set it as described in the description,

Each cluster is assigned the same IP, so one VS is created in the F5 BIG-IP Appliance. When you enter the kubectl get svc command in each cluster, it is confirmed that both clusters are assigned the same IP.

Even if you check in Infoblox IPAM, it is confirmed that only 1 IP is assigned.

#### Solution Proposed
I would like to be assigned different IPs from Infoblox IPAM through a separate identifier such as a key or ID.

#### Alternatives
All you need to do is to assign different IPs through the separator that was an option in the helm used when linking ipam and a separate annotation in the LB Type.

nandakishorepeddi:
**Description**:  Add MultiCluster support for IPAM controller by prepending custom keys

**Changes Proposed in PR**:

**Fixes**: resolves #3160 

## General Checklist
- [x] Smoke testing completed

## CRD Checklist


vklohiya (approved):
LGTM

Issue #3209, opened 2023-12-22
[LOGS] Missing context in some log messages


alonsocamaro:
#### Setup Details
 CIS Version : 2.15.1 (pre-release)
 Build: quay.io/f5networks/k8s-bigip-ctlr-devel:d3ba34519282070bc0e16778828d6bb28e63304d     
 
 #### Description  
 
The following log messages don´t have context required to figure out what/where it is missing:

Example 1) 

```
"2023/12/22 09:35:02 [INFO] [MultiCluster] There is no externalClustersConfig section or there are no clusters defined in it."
```

it should specify the resource type and resource name

Example 2) 

```
"2023/12/22 12:07:01 [WARNING] service '{service-green  openshift-ingress}'  not found"
```

It should indicate the cluster name

alonsocamaro:
Another example is the next similar log message:

```
2023/12/26 10:17:04 [ERROR] Error: Service openshift-ingress/service-green not found
```

Issue #3207, opened 2023-12-22
[COSMETIC] typo in log message Some of the mandatory parameters (clusterName/namespace/serviceName/servicePort) are missing


alonsocamaro:
#### Setup Details
 CIS Version : 2.15.1 (pre-release)     
 Build: quay.io/f5networks/k8s-bigip-ctlr-devel:d3ba34519282070bc0e16778828d6bb28e63304d        

#### Description  
  
The log message "Some of the mandatory parameters (clusterName/namespace/serviceName/servicePort) are missing" should say "port" instead of "servicePort" as per CRD

#### Steps To Reproduce

Configure a VS with extendedServiceReference and using "servicePort"

#### Expected Result 

The log should say "port" instead of "servicePort"

PR #3211, opened 2023-12-27
Improved the context for few logs

### LINKED ISSUE #3210: [LOGS] suggested additional changes
#### Title
It is believed the following could be made more user friendly

#### Description

About CIS HA, the following messages are shown when the primary goes down:

```
[cloud-user@ocp-provisioner cis-config]$ oc -n cis-mc-twotier logs deployment/k8s-bigip1-ctlr-deployment | grep primary
2023/12/26 10:16:48 [INFO] [MultiCluster] Enqueueing primary CIS/cluster down event
2023/12/26 10:16:48 [INFO] posting declaration on primary cluster down event
```

I find that:

1) the user would like to see when a when a probe fails as WARNING (indicating as suplementary the error message, no tcpdump should be required to figure out)
2) the user would like to see an additional message alike:
2023/12/26 10:16:48 [WARNING] [MultiCluster] X probes failed with primary CIS endpoint. Declaring the primary CIS DOWN.
3) if the primary comes back up an additional INFO message should be shown

The two messages above in the grep command could be DEBUG level IMHO

Additionally, It is not obvious to me (I don´t know what it means either) the following message:

```
2023/12/26 10:16:48 [INFO] [Request: 3] cluster local requested  in HACIS /
```

vklohiya:
**Description**: Improved the context for few logs

**Fixes**: resolves #3210 
#3209 
#3207 

## General Checklist
- [x] Smoke testing completed

### DIFF
--- a/pkg/controller/informers.go
+++ b/pkg/controller/informers.go
@@ -1101,6 +1101,8 @@ func getClusterLog(clusterName string) string {
 	clusterNameLog := ""
 	if clusterName != "" {
 		clusterNameLog = "from cluster: " + clusterName
+	} else {
+		clusterNameLog = fmt.Sprintf("from cluster: %v", Local)
 	}
 	return clusterNameLog
 }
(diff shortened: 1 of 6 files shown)


nandakishorepeddi (approved):
LGTM

PR #3224, opened 2024-01-08
#3208 service and servicePort params are required in Virtual Server CRD

### LINKED ISSUE #3208: [MINOR] Using only extendedServiceReferences fails
#### Setup Details
 CIS Version : 2.15.1 (pre-release)     
 Build: quay.io/f5networks/k8s-bigip-ctlr-devel:d3ba34519282070bc0e16778828d6bb28e63304d 

#### Description  
  
As per conversation with Vivek it should be possible to specify the following

```
apiVersion: "cis.f5.com/v1"
kind: VirtualServer
metadata:
  name: route-blue
  namespace: openshift-ingress
  labels:
    f5cr: "true"
spec:
  host: shop.multicluster.com
  ipamLabel: test
  hostGroup: multicluster.com
  tlsProfileName: tls-reencrypt-shop
  pools:
  - path: /
    monitor:
      interval: 5
      timeout: 16
      type: https
      name: /Common/shop.multicluster.com
      reference: bigip
    extendedServiceReferences:
    - clusterName: ocp1
      namespace: openshift-ingress
      serviceName: service-blue
      port: 443
    - clusterName: ocp2
      namespace: openshift-ingress
      serviceName: service-blue
      port: 443
    - clusterName: ocp3
      namespace: openshift-ingress
      serviceName: service-blue
      port: 443
```

where there is no default service stanza (which is used for the CIS HA instances) and instead all service instances in the different clusters should be specified in the extendedServiceReferences explicitly.

When the above configuration is used the AS3 declaration has missing the serviceName and hence the name of the pool doesn´t follow the expected syntax:

```
2023/12/22 11:47:36 [ERROR] [AS3] Raw response from Big-IP: map[code:422 declarationFullId: errors:[/mc-twotier/Shared: propertyName "__openshift_ingress_shop_multicluster_com" should match pattern "^[A-Za-z]([0-
9A-Za-z_.-]{0,188}[0-9A-Za-z_.])?$"] message:declaration is invalid] 
```

#### Steps To Reproduce
1) Apply the above configuration

#### Expected Result 

It should work as per converation with Vivek

#### Observations (if any)

- The behaviour indicated by Vivek is not documented at present and I find it should be documented, given that this alternative syntax (only using extendedServiceReferences) makes the VS definition independent on whether there is CIS in HA mode or not and in which CIS the VS config is applied (same manifest always)

- It is likely this also happens for the TransportServer

vidyasagar-m:
**Description**:  Using only extendedServiceReferences fails

**Changes Proposed in PR**: Made the service and servicePort params mandatory in the pool components of VS CRD

**Fixes**: resolves #3208

## General Checklist
- [ ] Added examples for new feature
- [ ] Updated the documentation
- [ ] Smoke testing completed

## CRD Checklist


vklohiya (approved):
LGTM

PR #3254, opened 2024-01-29
support sslProfile in https monitor

### LINKED ISSUE #3236: RFE: Support for automatic creation of https monitors with SNI
#### Title
Support for automatic creation of https monitors with SNI

#### Description

At time of this writing it is required for customers to create https monitors in order to health check an application that is hosting several domain names. This is the typical case of sending traffic to an ingress controller (2-tier deployments).

This requires:

- the creation of a server TLS profile
- the creation of the monitor itself

as follows:

```
ltm profile server-ssl serverssl.www.mc-nginx.com {
    app-service none
    defaults-from serverssl
    server-name www.mc-nginx.com
}
ltm monitor https www.mc-nginx.com {
    defaults-from https
    send "GET / HTTP/1.1\r\nHost: www.mc-nginx.com\r\nConnection: close\r\n\r\n"
    ssl-profile /Common/serverssl.www.mc-nginx.com
}
```

#### Actual Problem

Customers cannot automate the monitoring for this typical use case because the monitors need to be pre-created

#### Solution Proposed

When indicating an https monitor in the manifest, this should be by default SNI enabled

#### Additional context

In order for SNI enabled monitors to work in BIG-IP Classic it is necessary to enable the following DB key

```
tmsh modify sys db bigd.tmm value enable
```

See https://my.f5.com/manage/s/article/K85210713 for more details. Given that in-TMM monitoring has been affected by issues in the past and has an impact in data plane's performance. I would recommend that https monitors with SNI (which require in-TMM right now) have a different name than https in the manifests, for example: https-sni

lavanya-f5:
**Description**:  support sslProfile in https monitor

**Fixes**: resolves #3236 

## General Checklist
- [x] Smoke testing completed

## CRD Checklist

### DIFF
--- a/config/apis/cis/v1/types.go
+++ b/config/apis/cis/v1/types.go
@@ -148,6 +148,7 @@ type Monitor struct {
 	TargetPort int32  `json:"targetPort"`
 	Name       string `json:"name,omitempty"`
 	Reference  string `json:"reference,omitempty"`
+	SSLProfile string `json:"sslProfile,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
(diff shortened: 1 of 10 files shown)


vklohiya (approved):
LGTM

PR #3259, opened 2024-01-30
fix default value of sslProfile to none

lavanya-f5:
**Description**:  fix default value of sslProfile


nandakishorepeddi (approved):
LGTM

PR #3274, opened 2024-02-08
Add static routing mode to network config

lavanya-f5:
**Description**:  Add static routing mode to network config. Add description for deployconfig cr fields


nandakishorepeddi (approved):
LGTM

PR #3278, opened 2024-02-09
Multi host support for VS CRD

### LINKED ISSUE #3262: RFE: Add support for hosts array in VirtualServer CRD schemas (instead of just single Host string)
#### Title
Add support for hosts array in VirtualServer CRD schema.

#### Description
The current CRD schema for VirtualServer only supports a single Host string.  This is added as a match condition in the Traffic Policy rule.  This RFE is requesting support for a Hosts array to allow multiple hostnames to be specified to be included in the default forwarding rules/policies attached to the virtual server definition.

#### Actual Problem
I need to support multiple hosts and not just a single Host value in the match HTTP header is nay of <hosts> at request time default rule.  This is for our Disaster Recovery/failover plan, which requires the VirtualServer to forward both the primary hostname/domain and the DR hostname/domain to the same service/pool members.

Example:

www.customservice.com AND
www.dr.customservice.com

#### Solution Proposed
In addition to the existing Host string included in the spec, add an optional Hosts property of type array.  If specified, use this to add multiple hostnames to forwarding policy rules to match on multiple hosts within a single CRD VirtualServer definition.

#### Alternatives
N/A

#### Additional context
N/A

arzzon:
**Description**:  Multi host support for VS CRD
**Changes Proposed in PR**: Added hostAliases field where users can specify multiple host aliases for a Virtual Server CR.

**Fixes**: resolves #3262

## General Checklist
- [x] Smoke testing completed

## CRD Checklist

### DIFF
--- a/config/apis/cis/v1/types.go
+++ b/config/apis/cis/v1/types.go
@@ -28,6 +28,7 @@ type VirtualServerStatus struct {
 // VirtualServerSpec is the spec of the VirtualServer resource.
 type VirtualServerSpec struct {
 	Host                             string           `json:"host,omitempty"`
+	HostAliases                      []string         `json:"hostAliases,omitempty"`
 	HostGroup                        string           `json:"hostGroup,omitempty"`
 	HostGroupVirtualServerName       string           `json:"hostGroupVirtualServerName,omitempty"`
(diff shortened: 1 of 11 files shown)


nandakishorepeddi (approved):
LGTM

vklohiya (approved):
LGTM

PR #3289, opened 2024-02-15
Fix Improve log when admitting route status

vklohiya:
**Description**:  Fix Improve log when admitting route status

## General Checklist
- [x] Smoke testing completed


arzzon (approved):
LGTM

PR #3296, opened 2024-02-20
Improved log for tls handling in routes

### LINKED ISSUE #3264: RFE: Improve log 2024/01/31 09:41:47 [DEBUG] Updated Route xxx with TLSProfile
#### Title
RFE: Improve log 2024/01/31 09:41:47 [DEBUG] Updated Route xxx with TLSProfile

#### Description

The log above doesn´t indicate the TLSProfile being applied

#### Solution Proposed

Include the name of the profile. It would be a plus if it is possible to indicate from where it is getting that config, ie: baseRouteSpec? extendedRouteSpec's policyCR? from the manifest annotation? from the spec?

Given the non trivial nature of the rules governing SSL certificates used any hints in the logs would be appreciated

Right now I´m trying to figure out why certificates from the spec are not used and I don´t see any hint in the logs.

Please also consider showing which certs have been discarded and why

vklohiya:
**Description**:  Improved log for tls handling in routes

**Fixes**: resolves #3264 

## General Checklist
- [ ] Smoke testing completed

### DIFF
--- a/pkg/controller/nativeResourceWorker.go
+++ b/pkg/controller/nativeResourceWorker.go
@@ -161,7 +161,7 @@ func (ctlr *Controller) processRoutes(routeGroup string, triggerDelete bool) err
 					break
 				}
 
-				log.Debugf("Updated Route %s with TLSProfile", rt.ObjectMeta.Name)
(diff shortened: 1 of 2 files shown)


charanm08 (approved):
LGTM

PR #3308, opened 2024-02-26
support for setting default pool via policy CR

### LINKED ISSUE #3295: NextGenRoute Controller to support TLS passthrough without SNI in client hello
#### Title
NextGenRoute Controller to support TLS passthrough without SNI in client hello

#### Description
When an OpenShift Route is configured with `spec.tls.termination: passthrough`, NextGenRoute Controller creates a VS with an iRule that load balances traffic to different pools based on the SNI presented in the TLS Client Hello message. The VS does NOT have a default pool defined.

For clients which do not provide SNI in Client Hello, no pool can be selected, hence traffic is not passed to the pods in the OpenShift cluster.

#### Actual Problem
For applications hosting their own TLS certificates within the OpenShift clusters, clients which do not provide SNI in Client Hello are unable to access said applications in TLS passthrough mode.

#### Solution Proposed
Some toggle in the Route manifest to configure VS with pool member set to the backend pods without the need to select pools based on SNI in the iRules.

#### Alternatives
One suggestion is to remove the existing iRule with a Policy CR, but without a default pool defined, BIG-IP does not know where to send traffic to.

image

Maybe a way to configure default pool on the VS will be good.

#### Additional context
Extended Route spec
```
apiVersion: v1
data:
  extendedSpec: |
    extendedRouteSpec:
    - namespace: app
      vserverAddr: 10.1.10.200
      vserverName: override-vs
      policyCR: app/my-irules
kind: ConfigMap
metadata:
  labels:
    f5nr: "true"
  name: cis-extended-spec-override
  namespace: app
```

Policy to remove iRule from the VS
```
apiVersion: cis.f5.com/v1
kind: Policy
metadata:
  annotations:
  labels:
    f5cr: "true"
  name: my-irules
  namespace: app
spec:
  iRules:
    secure: none
```

lavanya-f5:
**Description**:  support for setting default pool via policy CR

**Changes Proposed in PR**:

**Fixes**: resolves #3295 

## General Checklist
- [x] Smoke testing completed

## CRD Checklist

### DIFF
--- a/config/apis/cis/v1/types.go
+++ b/config/apis/cis/v1/types.go
@@ -376,6 +376,7 @@ type PolicySpec struct {
 	SNAT         string           `json:"snat,omitempty"`
 	AutoLastHop  string           `json:"autoLastHop,omitempty"`
 	PoolSettings PoolSettingsSpec `json:"poolSettings,omitempty"`
+	DefaultPool  DefaultPool      `json:"defaultPool,omitempty"`
 }
 
 type PoolSettingsSpec struct {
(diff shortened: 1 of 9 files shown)


nandakishorepeddi (approved):
LGTM

PR #3317, opened 2024-02-29
Fix for missing rules for VS with HostAliases and hostGroup

arzzon:
**Description**:  Fix for missing rules for VS with HostAliases and hostGroup

**Changes Proposed in PR**: In case of VS with HostAliases and hostGroup defined the rule names will have hostGroup name as well as the host name for which this rule is meant to. This will ensure that rule name for all the host aliases are unique in case hostGroup is used.

## General Checklist
- [x] Smoke testing completed


vklohiya (approved):
LGTM

PR #3321, opened 2024-03-06
#3279 disable default partition

### LINKED ISSUE #3279: RFE: Allow default CIS partition to be disabled
#### Title
Allow default CIS partition to be disabled

#### Description
Add a feature flag to the CIS startup options to allow the default partition to be disabled (or make the `bigip-partition` configuration parameter optional).

#### Actual Problem
When exclusively using ConfigMap for CIS, the default partition created by CIS is superfluous. This poses usability and scalability issues in a large environment when multiple clusters are targeting the same BIG-IP device group. 

In such a case, each cluster has to be configured with a unique partition name, or else manage concurrency/contention issues if using a single shared partition. Additionally, the number of empty default partitions can become a usability and management problem on the BIG-IP device, both through the GUI and the CLI.

#### Solution Proposed
 Allow CIS administrators to disable the default partition created/used by CIS. This should be an option and not the default. 

Option 1:
Make the `--bigip-partition` configuration parameter optional; if it is omitted from the CIS startup arguments, do not create a default partition.

Option 2: 
Add a new `--disable-default-partition` configuration parameter, defaulting to false. If set to true, do not create a default partition. If set to true and `--bigip-partition` is also set, throw a startup error.

In either of these cases, it would potentially require additional validation on other startup parameters and state of other resources in the cluster; for example, CRDs would then need to ensure that a partition name is specified or else have CIS skip/error on such.

#### Alternatives
We are currently utilizing the name of the Kubernetes cluster as part of the default partition name and embedding that into our automation that deploys CIS. However, this is causing the usability problem mentioned above.

#### Additional context
Our company requires fine-grained control of the BIG-IP configuration and is also using AS3 for configuration outside of Kubernetes. As such, we have chosen to exclusively use ConfigMap as this provides us a single stream of configuration templates. Additionally, as infrastructure operators for our BIG-IP devices, we disable CRDs and Ingress - application teams will utilize our IaC CI/CD AS3 pipeline as a trusted pathway for security and regulatory-compliant load-balancer configuration.

vidyasagar-m:
**Description**:  #3279 default partition to be disabled

**Fixes**: resolves #3279

## General Checklist

### DIFF
--- a/cmd/k8s-bigip-ctlr/main.go
+++ b/cmd/k8s-bigip-ctlr/main.go
@@ -177,7 +177,7 @@ var (
 	shareNodes              *bool
 	overriderAS3CfgmapName  *string
 	filterTenants           *bool
-	disableDefaultPartition *bool
+	disableDefaultPartition bool
 
 	vxlanMode        string
 	openshiftSDNName *string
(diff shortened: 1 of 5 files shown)


vklohiya (approved):
LGTM

PR #3359, opened 2024-03-29
Fix for traffic not working in cluster mode

vklohiya:
**Description**:  Fix for traffic not working in cluster mode

## General Checklist
- [x] Smoke testing completed


charanm08 (approved):
LGTM

Issue #3393, opened 2024-04-26
Hub mode performance improve: use watch-list to replace regular query all service one by one


kylinsoong:
#### Hub mode performance improve

Currently Hub mode need more than 30 seconds to synchronize a POD IP change to BIG-IP, need increase performance.

#### Description

Current CIS Hub mode query all service(configured in configmap) one by one every 30 seconds, which means a service change may need long time to synchronize to BIG-IP.

~~~
        if appMgr.hubMode {
                // Leaving the old way for hubMode for now.
                svcListOptions := metav1.ListOptions{
                        LabelSelector: selector,
                }
        
                var err error
                // Identify services that matches the given label
                var services *v1.ServiceList
        
                services, err = appMgr.kubeClient.CoreV1().Services(v1.NamespaceAll).List(context.TODO(), svcListOptions)
        
                if err != nil {
                        log.Errorf("[CORE] Error getting service list. %v", err)
                        return nil, err
                }
                svcItems = services.Items
~~~

Too many API query can cause throttling request, decrease the whole system reliable, k8s client go watch-list and cache sdk are wide used to increase the reliability, CIS not hub mode already use this mechenism.

So I request use watch-list to improve hub mode performance. what the initial thoughts are:

* create a informer watch the service per namespace(Fortinet has a similiar controller, and keep firewall policy in hub namespace)
* create namespace informer to watch service in a fine-grained way, which only watch the namespace that are referenced in hub configmap

trinaths:
Created [CONTCNTR-4719] for internal tracking.

kylinsoong:
Hi @trinaths, after some research, this performance not only for Hub mode, but also for all configmap mode.

The appManager has a `agentCfgMap ` which contains all configmap in all namespace, each time the  appManager deploy resource, all configmap be send to Channel,

~~~
for _, cm := range appMgr.agentCfgMap {
	agentCfgMapLst = append(agentCfgMapLst, cm)
}
~~~

After the asManager received the request message from Channel, then it *iterator all tenants from all configmaps, for each tenants related services, the asManager will execute a query endpoints opeartion* this is the reason why syn POD IP changes to BIG-IP is so slowly once the configmap related services is large.

For no-hub mode configmap, the time spend to syn a POD change is:

~~~
Time_as3_update + Time_query_endpoint * Number_of_services
~~~

For hub mode configmap, the time spend to syn a POD change is:

~~~
[...]

PR #3435, opened 2024-05-24
#3393: hub mode support watch-list

kylinsoong:
This commit can increase CIS performance more than 95%, the key conclusion from performance test:

- If CIS watched services large than 200, POD scaling up/down, Add new service, delete exist service's performance will increase 95%
- The time of POD scaling up/down, Add new service, delete exist service is keep same dimension no matter how many services already exist. 

150 svc in 15 ns

100 svc in 10 ns

50 svc in 5 ns


vklohiya:
@kylinsoong , Please take care of review comments.

vklohiya:
Handled the changes in the PR: https://github.com/F5Networks/k8s-bigip-ctlr/pull/3437 with new label support `isTenantNameServiceNamespace: "true"`

PR #3477, opened 2024-07-08
Support for loadBalancerClass for service type lb

### LINKED ISSUE #3471: Support LoadBalancerClass
#### Title
Support LoadBalancerClass

#### Description
It is possible for clusters to have multiple implementations for `Service` of `type: LoadBalancer`. Kubernetes 1.24 and later have introduced the standard `.spec.loadBalancerClass` field in the service spec to be able to distinguish between the types of load balancing services available to the cluster, so that you can specify which load balancing class you would like to use. See [here](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class). I do not see any indication that this is presently supported in F5 CIS.

#### Actual Problem
If you have multiple load balancing classes, F5 CIS will not respect the `.spec.loadBalancerClass` field.

#### Solution Proposed
F5 CIS should have a config parameter to only watch services with the `loadBalancerClass` as specified in the param, and should default to the existing behavior of watching all load balancer classes.

vklohiya:
**Description**:  _Add Issue/Feature description_

**Changes Proposed in PR**:

**Fixes**: resolves #3471 

## General Checklist
- [x] Smoke testing completed

## CRD Checklist


arzzon (approved):
LGTM

PR #3488, opened 2024-07-17
Support for ipam-namespace deployment parameter

### LINKED ISSUE #3430: RFE: usability improvement add --ipam-namespace even it is yet not customizable to provide sane defaults
#### Title
RFE: usability improvement add --ipam-namespace even it is yet not customizable to we have sane defaults

#### Description

This RFE is to improve usability by having a sane default when IPAM is enabled. When using defaults in the components, these should be compatible.

#### Actual Problem

When using --ipam=true is used and not using --namespace=kube-system is a configuration error.

It is expected that IPAM namespace is to be configured. Although the IPAM namespace is not yet customizable, if we already provide a --ipam-namespace parameter with default set to kube-system we would provide a sane default when customers enable --ipam=true.

Additionally by using --ipam-namespace and not --namespace we don´t force the customers of not being able to use the --namespace-label which is incompatible with --namespace.

#### Solution Proposed

Add -ipam-namespace with defaults to kube-system **and** do not allow this setting to be modified until FIC allows this namespace to be customized. Throw an ERROR log message in such case.

vklohiya:
**Description**:  Support for ipam-namespace deployment parameter

**Changes Proposed in PR**:

**Fixes**: resolves #3430

## General Checklist
- [x] Smoke testing completed

### DIFF
--- a/cmd/k8s-bigip-ctlr/main.go
+++ b/cmd/k8s-bigip-ctlr/main.go
@@ -165,6 +165,7 @@ var (
 	sslInsecure                 *bool
 	ipam                        *bool
 	ipamClusterLabel            *string
+	ipamNamespace               *string
 	enableTLS                   *string
 	tls13CipherGroupReference   *string
 	ciphers                     *string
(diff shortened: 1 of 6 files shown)


charanm08 (approved):
LGTM

arzzon (approved):
LGTM

PR #3551, opened 2024-09-04
update edns pool members as per the hosts of the resource

vidyasagar-m:
**Description**:  CIS not removing pool members of GTM when host annotation removed on Transport server, Ingress link and service type LB

**Changes Proposed in PR**: Added host param to the enqueueupdate functions for delete event

## General Checklist
- [x] Smoke testing completed


vklohiya (approved):
LGTM

PR #3562, opened 2024-09-19
Fix for CIS with oneconnect and TLS breaks some connections

### LINKED ISSUE #3501: CIS with oneconnect and TLS breaks some connections
#### Setup Details
 CIS Version : 2.17.1   
 BIGIP Version: BIG-IP 17.1.0.1 Build 0.0.4 Point Release 1
 AS3 Version: f5-appsvcs-3.45.0-5.noarch
 Agent Mode: AS3
 Orchestration: OSCP  
 Orchestration 4.12
 Pool Mode: Nodeport      
 Additional Setup details: using HA Proxy as 2nd tier

#### Description  

When sending HTTPS requests to a re-encrypting VS. The serverside frequently sends a TLSv1.2 Record Layer: Alert (Level: Warning, Description: Close Notify). This is unexpected. 

After the Close Notify the BIG-IP sends an HTTP request within the same connection (which doesn´t make sense because we have notified the TLS connection is shut). The pool member obviously doesn´t send an HTTP reply and closes the connection at TLS level with another TLSv1.2 Record Layer: Alert (Level: Warning, Description: Close Notify), finally the TCP connection is shutdown.

In the client side this error is seen as follows:

curl: (52) Empty reply from server

The above behaviour is shown next (from file oneconnect-tcp9-tcp13.pcap which has clientside and serverside)

Screenshot 2024-07-26 at 14 55 33

The virtual server used in the test is /mc-twotier/Shared/crd_10_1_10_106_443 which has attached an iRule and an LTM profile. If I disable its irule (named /mc-twotier/Shared/crd_10_1_10_106_443_tls_irule), then oneconnect works fine. 

#### Steps To Reproduce

I have tested it with multi-cluster active-active but I believe this likely happens with single cluster configurations as well.

Configure a VS with a configuration alike next:
[vs.yaml.txt](https://github.com/user-attachments/files/16392188/vs.yaml.txt)
[tlsprofile.yaml.txt](https://github.com/user-attachments/files/16392189/tlsprofile.yaml.txt)

Run a command as follows:

while true ; do curl https://www.demo-mc.com/shop ; done

#### Expected Result 

The server side connection shouldn´t break. Connection breaks because the BIG-IP sends a TLS Close Notify.

#### Actual Result  

BIG-IP closes the connection at TLS level before the server sends a reply.

#### Diagnostic Information

Sharepoint link with complementary data shared with Vivek and Trinath

vklohiya:
**Description**:  Fix for CIS with oneconnect and TLS breaks some connections

**Fixes**: resolves #3501 

## General Checklist
- [x] Smoke testing completed

### DIFF
--- a/docs/RELEASE-NOTES.rst
+++ b/docs/RELEASE-NOTES.rst
@@ -15,6 +15,7 @@ Bug Fixes
 * `Issue 3520 <https://github.com/F5Networks/k8s-bigip-ctlr/issues/3520>`_: [LOGS] improve log message Finished syncing virtual servers xxx in namespace yyy(199.218µs), 1/7
 * `Issue 3542 <https://github.com/F5Networks/k8s-bigip-ctlr/issues/3542>`_: Change of weight from primary to alternate backend is not graceful for Transport server
 * `Issue 3507 <https://github.com/F5Networks/k8s-bigip-ctlr/issues/3507>`_: Controller stops posting changes at runtime when multiple ingress use the same backend service
+* `Issue 3501 <https://github.com/F5Networks/k8s-bigip-ctlr/issues/3501>`_: CIS with oneconnect and TLS breaks some connections
 * Remove pool members of GTM when host removed or updated on transport server, ingresslink, and, service type lb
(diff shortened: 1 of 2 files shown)


charanm08 (approved):
LGTM

PR #3642, opened 2024-11-08
#3523 suport for http compression profile in VS CRD

### LINKED ISSUE #3523: HTTP Compression Profile in CRD
#### Title
Setting HTTP Compression Profile in CRD

#### Description
We use http compression profile in our virtual server. We are looking for a way to set HTTP Compression Profile in CRD. 

#### Actual Problem
We could not find a way to set HTTP compression Profile in CRD. I know there is a way with config maps, or I can create a iRule to accomplish it.

#### Solution Proposed
Adding a field while setting a virtual server in CRD can solve the problem.

#### Alternatives
If there is an iRule that performs the same function as HTTP compression and WAN-optimized compression, that will suffice for now.

vidyasagar-m:
**Description**:  HTTP Compression profile in VS and Policy CRD

**Changes Proposed in PR**: Added support for HTTP compression profile for Virtual Server and Policy CR

**Fixes**: resolves #3523 

## General Checklist
- [x] Smoke testing completed

## CRD Checklist


vklohiya (approved):
LGTM

PR #3668, opened 2024-12-03
fix implicit service discovery standalone mode

lavanya-f5:
**Description**:  fix implicit service discovery standalone mode


vklohiya (approved):
LGTM

PR #3692, opened 2024-12-20
fix target port handling for ts ab datagroup

lavanya-f5:
**Description**:  fix target port handling for ts ab datagroup


vklohiya (approved):
LGTM

PR #3802, opened 2025-04-08
feat: support custom CRD label

### LINKED ISSUE #3559: Support integrating with multiple F5 instances from a single Kubernetes cluster
#### Title
Support integrating with multiple F5 instances from a single Kubernetes cluster

#### Description
In our setup we have a Kubernetes cluster that needs to integrate with two F5 instances. One F5 is publicly exposed and serves frontend traffic, the other F5 is only used internally aka the backend traffic.

#### Actual Problem
We integrated two F5 instances into a single cluster using the [ingress configuration of the CIS](https://clouddocs.f5.com/containers/latest/userguide/ingress.html). We have two CIS controller pods running with each their own dedicated `ingressClass` definition. Using this we can target the ingress for either the frontend or backend. 

This setup works but the integration of type `ingress` is very limited in its functionality.
- You cannot configure a different heath-check target port compared to the service port
- SSL profiles cannot be customised
- ...

Looking at the documentation for the other integration options it seems like the CIS was not designed to work with multiple F5 instances? Ideally this is build into the CRDs.

#### Solution Proposed
Support the configuration of multiple F5 instances from a single cluster or if it is already supported, document the recommended configuration options etc.

#### Alternatives
We came across a new class identifier implementation for the service type load balancer: https://clouddocs.f5.com/containers/latest/userguide/loadbalancer/#load-balancer-class-support

From the wiki:
> Load Balancer Class is supported for all the Custom Resources (VirtualServer, TransportServer and IngressLink) and loadBalancer service by default and can not be disabled.

This sounds like we can set a `loadBalancerClass` property on all those custom resources and the CIS will monitor only those instances matching its class configuration? And because it can be set on the CRDs it means all configuration options of the CRD integration configuration become available?

arzzon:
**Description**:  Support custom CRD label

**Changes Proposed in PR**: support custom CRD label

**Fixes**: resolves #3559 

## General Checklist
- [x] Smoke testing completed

## CRD Checklist

### DIFF
--- a/cmd/k8s-bigip-ctlr/main.go
+++ b/cmd/k8s-bigip-ctlr/main.go
@@ -169,6 +169,7 @@ var (
 	manageIngressClassOnly      *bool
 	ingressClass                *string
 	loadBalancerClass           *string
+	customResourceLabel         *string
 	manageLoadBalancerClassOnly *bool
 	bigIPURL                    *string
 	bigIPUsername               *string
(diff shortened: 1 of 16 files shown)


vklohiya (approved):
LGTM

PR #3815, opened 2025-04-21
fix: tenant cache to avoid the concurrent read write

vklohiya:
**Description**:  There was a corner case where Agent is reading the cached tenants before it's update with the current BIGIP config, Hence some time it was reporting that there is no changes made to the config.

## General Checklist
- [x] Smoke testing completed


charanm08 (approved):
LGTM

lavanya-f5 (approved):
LGTM

PR #3822, opened 2025-04-23
add crd validation to reject partition update

lavanya-f5:
**Description**:  add crd validation to reject partition update

## CRD Checklist


vklohiya (approved):
LGTM

PR #3861, opened 2025-06-03
il mc retry pool fix

lavanya-f5:
**Description**:  fix irule to retry pool on 503 with ingresslink multicluster app failover

**Changes Proposed in PR**: increase retry count for each pool (try twice before sending 503 to client) with ingresslink multicluster failover scenario to handle scenarios of pool transition 

## General Checklist
- [x] Smoke testing completed


vklohiya (approved):
LGTM

PR #3867, opened 2025-06-09
feat: adding the webhook support for the custom resources

### LINKED ISSUE #3754: can CIS or AS3 have an "ignore and continue" behaviour when receiving 422 errors?
#### Setup Details
 CIS Version : 2.19     
 BIGIP Version: Big IP 17.1.1 

#### Description  
Customer reports that when a duplicate IP address and port are accidentally configured in CRD's in K8s, the resulting 422 response from BIG-IP to CIS causes all new changes in K8s to fail to be processed by CIS.

Customer admits that creating 2x CRD's with the same IP:port is a mistake on their end, but these are mistakes from other teams that the F5 admin cannot control.

Is there a way for CIS to ignore CRD's that cause duplicate IP:port, or ignore the 422 response from BIG-IP, and continue processing remaining changes?

This is a problem for multiple customers: One mistake in a CRD causes CIS to effectively stop communicating changes to BIG-IP until it is fixed, but usually discovery of this mistake is difficult and many hours/days after it happens.

Basically, can CIS or AS3 have an "ignore and continue" behaviour?  

#### Steps To Reproduce
1) Create TransportServer with CRD with specific IP:port combination
2) Let CIS create VS on BIG-IP
3) Create another TS CRD with same IP:port combo. 
4) CIS will receive error from BIG-IP due to duplicate IP:port combo **and stop processing any new changes to K8s objects.**

#### Expected Result 
CIS could ignore errors caused by duplicate IP:ports on BIG-IP and continue to process new changes.

#### Actual Result  
CIS re-sends declaration, continuously receving 422 errors.

#### Diagnostic Information
```buildoutcfg
```

vklohiya:
**Description**:  As part of this PR we are adding the webhooks for CIS to support the custom resource validation and mutation.

**Changes Proposed in PR**:

Validating Webhook

The validating webhook is a type of [ValidatingAdmissionWebhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook) which runs against Custom Resources (CRs) at CREATE or UPDATE time to validate the CR contains a configuration which will yield an operable cluster. The CustomResourceDefinitions use OpenAPI schema validation for basic property checks (such as integer bounds and non-empty strings), but more complex validation of the configuration is done via the webhook.

Mutating Webhook

The Mutating webhook is a type of [MutatingAdmissionWebhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) which runs against Custom Resources (CRs) at CREATE or UPDATE time to populate default values in the CR's spec.

**Fixes**: resolves #3754 

## General Checklist
- [x] Smoke testing completed


lavanya-f5 (approved):
LGTM

PR #3870, opened 2025-06-11
fix: LB type service status update

arzzon:
**Description**:  Fix for LB type service status update

**Changes Proposed in PR**: Following code fix is done for LB type service status update
* If LB service is present as pool for VS/TS then while updating the status of LB service avoid taking the Virutal Address directly from VS/TS status for updating the LB type service status.

## General Checklist
- [x] Smoke testing completed

## CRD Checklist


vklohiya (approved):
LGTM

PR #3922, opened 2025-07-30
#3895 increase the failure threshold value to 60

### LINKED ISSUE #3895: Improve logging when BIG-IP is not reachable during pod initialization
#### Setup Details
 CIS Version : 2.20.0      
 Build: f5networks/k8s-bigip-ctlr:2.20.0     
 BIGIP Version: n/a      
 AS3 Version: n/a
 Agent Mode: AS3       
 Orchestration: K8S      
 Orchestration Version:       
 Pool Mode: Nodeport      
 Additional Setup details: <Platform/CNI Plugins/ cluster nodes/ etc>

#### Description  
When CIS is installed with a BIG-IP that cannot be reached, the pods are never healthy (failing liveness and readiness probes), and continue to `CrashLoopBackOff` until the restart limit is reached. The usability issue I am reporting is that there is never any message logged in the pod logs that is indicative of this condition.

#### Steps To Reproduce
1) Deploy CIS using manual or Helm instructions (it makes no difference), but specify a BIG-IP address that either does not exist, or is unreachable (for instance, blocked by a firewall rule). Example: specify an arg of `bigip-url=1.2.3.4` at install time.
2) Examine CIS pod logs
3) Note the lack of explanation causing the continuous pod restarts.

#### Expected Result 
A CRITICAL error message should be written to the log indicating that the BIG-IP could not be reached

#### Actual Result  
No log message indicating this condition, causing much troubleshooting and toil.

vidyasagar-m:
**Description**:  Improve logging when BIG-IP is not reachable during pod initialization

**Changes Proposed in PR**: Increased the failure threshold to 60 secs

**Fixes**: resolves #3895

## General Checklist
- [x] Smoke testing completed


vklohiya (approved):
LGTM

PR #3977, opened 2025-10-15
Avoid Deletion of unmanaged routes

lavanya-f5:
**Description**:   Avoid static route deletion for unmanaged routes in /Common partition


arzzon (approved):
LGTM

PR #4055, opened 2026-06-03
fix: link check ignore for broken issues backlink

lavanya-f5:
**Description**:  link check ignore for broken issues backlink


vidyasagar-m (approved):
LGTM

